PR #23, opened 2014-12-01
Container specific setup for kie(-drools)-wb

psiroky:
Please _don't_ merge this yet, it is still work in progress, but I figured I would share what I currently have to get some feedback.

Currently there are two assemblies: Tomcat7 and Tomcat8. They are mostly the same, unfortunately Tomcat8 made some small changes in `server.xml`. I wanted to keep the configs as much as possible in sync with the default ones, but maybe we could create just one assembly that would work on both tomcats.

Other thing is the naming. I could not come with better name that `kie-wb-container-setup`. I don't like it much, but can't think of anything better right now. Suggestions more than welcome.

I believe it could also be useful to add some predefined roles and users into `tomcat-users.xml`. WDYT?

@mswiderski when you few minutes could you take a look? This is the first step for the out-of-the-box config on Tomcat we talked about ~week ago.

### DIFF
--- a/kie-tomcat-integration/src/main/java/org/kie/integration/tomcat/JACCValve.java
+++ b/kie-tomcat-integration/src/main/java/org/kie/integration/tomcat/JACCValve.java
@@ -51,84 +51,82 @@ public class JACCValve extends ValveBase {
 
     private static final Logger logger = LoggerFactory.getLogger(JACCValve.class);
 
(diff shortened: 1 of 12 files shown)


psiroky:
Closing this as it is outdated. I will look into this in future and see what can be done directly in the WAR and what needs to be configured on the Tomcat side.

PR #26, opened 2015-01-14
BZ-1002303 Enable users to apply the locale of their choice for all UI's...

bxf12315:
... that have been localized


Salaboy:
merged in master

PR #29, opened 2015-02-04
[GUVNOR-2168] remove assemblies for JBossAS 7

psiroky:
Removing the JBoss AS 7 assemblies as the server is quite old and we already support the latest community version - WildFly 8.x.

This needs to be done only for master. For 6.2.x the assemblies should stay.

I talked with @krisv and he is OK with this change.

@etirelli @mdproctor please confirm that is OK also from you point of view.


etirelli:
+1

PR #30, opened 2015-02-23
REST with httpcomponents's fluent-hc and some updates to guvnor tests

mrietveld:
@psiroky This also fixes the EAP 6.4.x problems. 

There are problems with the war on the master branch, so I wasn't able to test it there. 

I've refactored the RestUtil to use the apache httpcomponents for a couple of reasons: 
- the KieRemoteHttpRequest object that I created doesn't make sense: I want to remove that in master (we're a bpm/rules project, not an HTTP project so no need to reinvent that wheel)
- I initially moved away from RestEasy because of OSGI problems. However, the RestEasy interface also changes with 3.0.x, which is what we then would have to use with wildfly/eap 7. And that's another pain to have to rewrite that code as well. 
- I also want to stay away from CXF for similar reasons. CXF also has a very "heavy" dependency tree for what we want, shich is simply doing HTTP calls. 

What I tried to do here was a couple things: 
- make the RestUtil interface framework agnostic: the tests don't need to import any code from any REST or HTTP frameworks
- move both Rest Tests to the same class and way to do REST operations
- within the RestUtil class itself, use framework that was 1. as lightweight as possible and 2. external 

 Let me know what you think, and if would be okay to merge this into 6.2.x and master.


psiroky:
I will take a look, but overall I agree with what you wrote above. 

One question: would it maybe make sense to keep the tests classes (for guvnor rest and kie-remote rest) separated, and just let them reuse the common code? This way you can easily run only one kind of tests by saying the class name. Useful when testing just kie-drools-wb which does not have the kie-remote capabilities (but than can also by done by the junit categories...).

kie-wb-smoke-tests/src/test/java/org/kie/smoke/wb/AbstractWorkbenchIntegrationTest.java
```diff
@@ -23,7 +23,7 @@
 
     static {
         try {
-            deploymentUrl = new URL(System.getProperty("deployable.base.uri"));
+            deploymentUrl = new URL(System.getProperty("deployable.base.uri", "http://localhost:8080/kie-wb/"));
```
psiroky: This is exactly what I have been doing locally when running the tests from IDE. However, this will only work for kie-wb as kie-drools-wb has different app context. 

To support both I see two options here:
 * make the URL detection kind of smart: try the /kie-wb and if that is not respoding, try /kie-drools-wb)
 * use the same app context for both kie-wb and kie-drools-wb (e.g. just "wb")

I would vote for 2nd option as is seems less tricky.

kie-wb-smoke-tests/src/test/java/org/kie/smoke/wb/rest/GuvnorRestSmokeIntegrationTest.java
```diff
@@ -251,65 +266,70 @@ public void testManipulatingRepositoriesAndProjects() throws Exception {
...
+                jobStatus.equals(JobStatus.ACCEPTED) || jobStatus.equals(JobStatus.APPROVED) );
         int wait = 0;
-        do {
-            Thread.sleep(jobCompleteSleepSecs * 1000);
-            ClientRequest restRequest = requestHelper.createRequest("jobs/" + jobId);
-            JobResult jobResult = get(restRequest, mediaType, JobResult.class);
-            logger.debug("]] " + convertObjectToJsonString(jobResult));
-            assertEquals(jobResult.getJobId(), jobId);
+        JobResult jobResult = null;
+        while( ! jobStatus.equals(JobStatus.SUCCESS) && wait < maxTries ) {
```
psiroky: In some cases the expected status could be different than the SUCCESS, for example when checking if the server correctly returns NOT_FOUND for resources that does not exists. 

Could we check here that the status changed from the initial APPROVED or ACCEPTED to some "final" state (e.g. the NOT_FOUND or SUCCESS).

kie-wb-smoke-tests/src/test/java/org/kie/smoke/wb/util/RestRepositoryDeploymentUtil.java
```diff
@@ -34,51 +40,75 @@
...
+        waitForDeploymentToFinish(sleepSecs, true, deployJob.getDeploymentUnit());    
     }
     
     // submethods ------------------------------------------------------------------------------------------------------------
+  
+    /**
+     * Delete the repository with the given repository name
+     * @param repositoryName
+     * @return A {@link JobRequest} instance returned by the request with the initial status of the request
+     */
```
psiroky: The javadoc seems to be outdated here.

kie-wb-smoke-tests/src/test/java/org/kie/smoke/wb/util/RestUtil.java
```diff
@@ -37,270 +48,359 @@
...
-     * Call the POST HTTP method on the given request and then releases the connection.
-     * @param restRequest The {@link ClientRequest} instance with the request information
-     * @param mediaType The {@link MediaType} that should be used with the request
-     * @param status The HTTP status that the request should return
-     * @return The actual entity returned by the request: this can be null
-     */
-    public static void post(ClientRequest restRequest, MediaType mediaType, int status) {
-        ClientResponse<?> responseObj = httpMethodReturnResponse(POST, restRequest, mediaType, status);
-        responseObj.releaseConnection();
+    public static void failAndLog( String msg, Exception e ) {
```
psiroky: Nit-picking here...but based on what the method does, I think it should be called `logAndFail()`

kie-wb-smoke-tests/src/test/java/org/kie/smoke/wb/AbstractWorkbenchIntegrationTest.java
mrietveld: Oops, this made it in by accident, but I agree with your second suggestion ("wb"). 

Will change it to that.

kie-wb-smoke-tests/src/test/java/org/kie/smoke/wb/rest/GuvnorRestSmokeIntegrationTest.java
mrietveld: Agreed, will change.

kie-wb-smoke-tests/src/test/java/org/kie/smoke/wb/util/RestRepositoryDeploymentUtil.java
mrietveld: oh, good point! Will add that.

kie-wb-smoke-tests/src/test/java/org/kie/smoke/wb/util/RestUtil.java
mrietveld: Haha.. okay, makes sense. Will do.

psiroky:
@mrietveld any news on this one? I would still very much like to see the tests pass on 6.2.x, so it would be great to get this in (and also cherry-pick to master).

mrietveld:
@psiroky Thanks for pinging me -- I've been a little swamped this week (and the weeks before that). I'll get to it next week!

kie-wb-smoke-tests/src/test/java/org/kie/smoke/wb/AbstractWorkbenchIntegrationTest.java
mrietveld: Removed from code (see new PR for master branch).

kie-wb-smoke-tests/src/test/java/org/kie/smoke/wb/rest/GuvnorRestSmokeIntegrationTest.java
mrietveld: Changed (see new PR for master branch). 

            while( ( jobStatus.equals(JobStatus.ACCEPTED) || jobStatus.equals(JobStatus.APPROVED) )
                && wait < maxTries ) {

kie-wb-smoke-tests/src/test/java/org/kie/smoke/wb/util/RestRepositoryDeploymentUtil.java
mrietveld: Fixed (see new PR for master branch).

kie-wb-smoke-tests/src/test/java/org/kie/smoke/wb/util/RestUtil.java
mrietveld: Changed (see new PR for master branch).

kie-wb-smoke-tests/src/test/java/org/kie/smoke/wb/AbstractWorkbenchIntegrationTest.java
psiroky: Marco, I can't see the new PR for master branch.

mrietveld:
Closing this because a new PR has been made for master: https://github.com/droolsjbpm/kie-wb-distributions/pull/46

PR #33, opened 2015-03-20
BZ-1203696: application of proper use of Roles and Groups

porcelli:


manstis:
http://github.com/droolsjbpm/kie-wb-distributions/commit/53c1ac903

PR #42, opened 2015-05-10
GWT upgrade - improvements for Weblogic and WebSphere assemblies

mswiderski:
@porcelli @manstis @dgutierr 
please have a look at the changes as these were required to successfully deploy and run on WLS and WAS.

See email for still remaining issues

### DIFF
--- a/kie-drools-wb/kie-drools-wb-distribution-wars-cdi1.0/pom.xml
+++ b/kie-drools-wb/kie-drools-wb-distribution-wars-cdi1.0/pom.xml
@@ -343,6 +343,11 @@
       <groupId>org.jbpm</groupId>
       <artifactId>jbpm-services-ejb-timer</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.codehaus.woodstox</groupId>
+      <artifactId>stax2-api</artifactId>
+      <version>3.1.1</version>
+    </dependency>
 
     <!-- Errai integration that supports CDI 1.0 -->
     <dependency>
--- a/kie-drools-wb/kie-drools-wb-distribution-wars-cdi1.0/src/main/assembly/weblogic12/WEB-INF/weblogic.xml
+++ b/kie-drools-wb/kie-drools-wb-distribution-wars-cdi1.0/src/main/assembly/weblogic12/WEB-INF/weblogic.xml
@@ -6,6 +6,7 @@
       <wls:package-name>com.google.common.*</wls:package-name>
       <wls:package-name>org.xmlpull.*</wls:package-name>
       <wls:package-name>org.apache.commons.io.*</wls:package-name>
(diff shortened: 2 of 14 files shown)


manstis:
Looks reasonable enough. Shouldn't the version number information however
be moved to one of droolsjbpm-build-bootstrap's boms/poms?

mswiderski:
yes, once we agree on this it should, same for errai version for cdi 1.0 I believe

kie-drools-wb/kie-drools-wb-distribution-wars-cdi1.0/pom.xml
```diff
@@ -343,6 +343,11 @@
       <groupId>org.jbpm</groupId>
       <artifactId>jbpm-services-ejb-timer</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.codehaus.woodstox</groupId>
+      <artifactId>stax2-api</artifactId>
+      <version>3.1.1</version>
+    </dependency>
```
psiroky: I thought we are enforcing that the artifact version needs to be in `dependencyManagement`. Looking at the POM, the enforcer rule checking this was disable which is I think bad. We need to enable it again and move this version into `dependencyManagement` section (or directly into the kie-p-w-d).

psiroky:
It seems that we can't move the `errai-weld-integration:3.0.4.Final` into `kie-p-w-d`, because there is already the same artifact being imported from `errai-internal-bom`, but in version `3.2.0-SNAPSHOT`. So would need to override the version in the `kie-wb-distribtuions-cdi1.0` (and in ideal case use the `dependencyManagement` section.

kie-drools-wb/kie-drools-wb-distribution-wars-cdi1.0/pom.xml
mswiderski: +1 on moving this to bom once agreed to merge this

Dnia 10 maj 2015 o godz. 13:40 Petr Široký <notifications@github.com> napisał(a):

porcelli: @psiroky as far as i know this is the only way to have multiple versions of a dependency for a single release execution (use of profiles would be an alternative, but it would need 2 different executions).
And enforcement are disable to be able to have such versioning.

psiroky: I think it should be possible to include `dependencyManagement` section directly in `kie-wb-distribution-wars-cdi1.0` module and define (override) the version there. Then the `dependencies` section can use the artifact without specifying version and it will get the correct one. Then we should also be able to re-enable the enforcer again.

I will look into this and submit a PR in case I make it work.

PR #45, opened 2015-05-12
JBPM-4625 Generation of JPA enabled models

wmedvede:


manstis:
http://github.com/droolsjbpm/kie-wb-distributions/commit/af056bab9

PR #46, opened 2015-05-13
REST with httpcomponents's fluent-hc and some updates to guvnor tests

mrietveld:
@psiroky This also fixes the EAP 6.4.x problems. 

There are problems with the war on the master branch, so I wasn't able to test it there. 

I've refactored the RestUtil to use the apache httpcomponents for a couple of reasons: 
- the KieRemoteHttpRequest object that I created doesn't make sense: I want to remove that in master (we're a bpm/rules project, not an HTTP project so no need to reinvent that wheel)
- I initially moved away from RestEasy because of OSGI problems. However, the RestEasy interface also changes with 3.0.x, which is what we then would have to use with wildfly/eap 7. And that's another pain to have to rewrite that code as well. 
- I also want to stay away from CXF for similar reasons. CXF also has a very "heavy" dependency tree for what we want, shich is simply doing HTTP calls. 

What I tried to do here was a couple things: 
- make the RestUtil interface framework agnostic: the tests don't need to import any code from any REST or HTTP frameworks
- move both Rest Tests to the same class and way to do REST operations
- within the RestUtil class itself, use framework that was 1. as lightweight as possible and 2. external


mrietveld:
@psiroky I've added a second commit that moves the utility classes to `src/main/java`. 

The reason for this is so that I can use the same utility classes in my development integration tests. That way, the transition from development (kie-tests) to stable (kie-wb-smoke-tests) integration tests is seamless, because I'll be using the exact same utility class in both. 

Does this sound okay?

psiroky:
Merged into master https://github.com/droolsjbpm/kie-wb-distributions/commit/47738d8e3e0c21044aa75071cb09a76776aec43d https://github.com/droolsjbpm/kie-wb-distributions/commit/56b7c050dbc3fe65d964130f7701a3be5ac2542d

PR #48, opened 2015-05-15
[JBPM-4555] merge kie-drools-wb war modules back into one

psiroky:
* the solution is using maven-dependency-plugin to copy
   CDI 1.0 errai-weld-integration into target/ and then
   includes that in the required distributions (EAP, Tomcat, WAS, WLS).
   The WildFly distro is not touched by this change

 * it should be easier to maintain the WAR configs when they
   are only in single module

 * it should be also easier for users as they don't have to
   know there is the cdi1.0 in the WAR names

 * this also simplifies testing, as we can rely on the fact
   that artifactId is always the same, and only classifer is
   different for each WAR

@porcelli @manstis @mswiderski please take a look and let me know what you think. This is only for `kie-drools-wb`. If we decide to use this, I will do the same for `kie-wb`.


psiroky:
Just to clarify a bit more, the PR is about merging `kie-drools-wb-distribution-wars` and `kie-drools-wb-distribution-wars-cdi1.0` back into single module `kie-drools-wb-distribution-wars`.

manstis:
+1 from me. 

With PR applied locally kie-drools-wb built OK. I could see the different errai-weld-integration  JARs in the respective WARs. Test deployed on EAP6.4 and Wildfly. Both webapps ran OK (only a very quick tyre-kick).

psiroky:
I did directory compare (`diff -qr`) before and after the changes and the resulting WARs were identical, so from this point of view the change is safe.

psiroky:
Merged into master for both `kie-drools-wb` and `kie-wb`. https://github.com/droolsjbpm/kie-wb-distributions/commit/4a9c680937b43bc2a7537f7bb5a79151489b0e23

PR #49, opened 2015-05-19
BPMSPL-160: DRL text editor improvements

manstis:
See https://github.com/droolsjbpm/drools-wb/pull/31


manstis:
There is a related PR on drools-wb too.

See https://github.com/droolsjbpm/drools-wb/pull/31

porcelli:
http://github.com/droolsjbpm/kie-wb-distributions/commit/5b2a9ac9c

PR #50, opened 2015-05-19
BZ-1221096: adding security headers for exec servers

porcelli:


manstis:
http://github.com/droolsjbpm/kie-wb-distributions/commit/03d45b1b8

PR #51, opened 2015-05-20
[JBPM-4555] reuse common configuration between community and product assemblies

psiroky:
* re-using the assembly config makes it easier to change things
   as there is less copy&paste needed

 * the community and product assemblies are almost identical, only
   few resources under src/main/producized are added into the
   product ones. The common component descriptor contains the
   overall config that can be directly used by the community
   assembly. The additional resources are added/replaced in the
   product assembly XML file.

* I compared the final WARs before and after the changes and they are identical

@manstis @porcelli @mswiderski please take look and let me know in case I missed something.


manstis:
+1 from me (I like the concept and approach. I've not checked your ability to copy and paste from the old files to the new!).

porcelli:
+100000000000000000000
Now why not try to remove kie-wb and kie-drools-wb and externalize UI setup using config files? So we'll get rid of most duplication

mswiderski:
if I got it right it looks like it's based on maven assembly components (that we talked sometime ago https://maven.apache.org/plugins/maven-assembly-plugin/examples/single/using-components.html) then here is my +10000 (as as many zeros you like ;))

great job!

manstis:
So who's going to merge this one then? ;)

mswiderski:
I can merge that in if it can wait until tomorrow, calling a day now :)

manstis:
http://github.com/droolsjbpm/kie-wb-distributions/commit/bf96b054c

PR #52, opened 2015-05-22
[BZ-1187211] remove dependency on private modules for EAP and WildFly distros

psiroky:
* removed the private/unsupported modules from jboss-deployment-structure.xml
   (as logged by WildFly and EAP during deployment)
 * jars from the private modules need to be directly in WEB-INF/lib
 * updated the other assemblies (includes/excludes) to reflect these changes
 * EAP/WildFly distros few additional jars in WEB-INF/lib
 * Tomcat/WAS/WLS distros are exactly the same as before
   (even though the assemblies changed)
 * there are two exceptions: dom4j and CXF. See the comments in the
   jboss-deployment-structure.xml files for more details.

@manstis @porcelli @mswiderski please take a look and let me know what you think.

### DIFF
--- a/kie-drools-wb/kie-drools-wb-distribution-wars/src/main/assembly/component-kie-drools-wb-weblogic-12-common.xml
+++ b/kie-drools-wb/kie-drools-wb-distribution-wars/src/main/assembly/component-kie-drools-wb-weblogic-12-common.xml
@@ -15,36 +15,23 @@
     <dependencySet>
       <includes>
         <include>antlr:antlr</include>
-        <include>commons-cli:commons-cli</include>
-        <include>commons-codec:commons-codec</include>
-        <include>commons-collections:commons-collections</include>
-        <include>commons-configuration:commons-configuration</include>
-        <include>commons-io:commons-io</include>
-        <include>commons-lang:commons-lang</include>
-        <include>org.apache.cxf:*</include>
         <include>dom4j:dom4j</include>
-        <include>org.apache.httpcomponents:httpclient</include>
+        <include>org.apache.cxf:*</include>
+        <include>org.apache.neethi:neethi</include>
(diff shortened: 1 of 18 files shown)


kie-drools-wb/kie-drools-wb-distribution-wars/src/main/assembly/component-kie-drools-wb-weblogic-12-common.xml
```diff
@@ -83,9 +70,12 @@
           <exclude>WEB-INF/lib/errai-jboss-as-support-*.jar</exclude>
           <!-- Errai 1.1+ CDI Compat -->
           <exclude>WEB-INF/lib/errai-weld-integration-*.jar</exclude>
+          <exclude>WEB-INF/lib/org.osgi.core-*.jar</exclude>
+          <exclude>WEB-INF/lib/javassist-*.jar</exclude>
+          <exclude>WEB-INF/lib/jgroups-*.jar</exclude>
```
psiroky: This is here to make sure the wars look the same after this change. TBH I am not quite sure it is correct. These jars are in EAP/WildFly wars, so either they are not needed there or they should be included in the WAS/WLS wars as well (as they are required by the apps to work correctly).

mswiderski:
+1
was just wondering what in our projects does require direct dependency to jgroups?

psiroky:
`errai-bus` depends on `jgroups`: https://github.com/errai/errai/blob/master/errai-bus/pom.xml#L206

psiroky:
But I agree that this is weird since we are not bundling it in WAS/WLS WARs. After we merge this one I'll look into these suspicious deps.

manstis:
http://github.com/droolsjbpm/kie-wb-distributions/commit/70c8ca789

PR #53, opened 2015-05-23
Override Errai DEBUG logging using own logback.xml config

psiroky:
* the root level is set to info, so only info messages
   will be printed

 * the logback.xml is also excluded from the final WAR
   as it should not be there (it would then need to be
   removed in the distribution-wars assemblies)

* as discusses with @csadilek this should be the best
  way of suppressing long Errai DEBUG output during
  gwt:compile

### DIFF
--- a/kie-drools-wb/kie-drools-wb-webapp/pom.xml
+++ b/kie-drools-wb/kie-drools-wb-webapp/pom.xml
@@ -1533,7 +1533,12 @@
       <plugin>
         <artifactId>maven-war-plugin</artifactId>
         <configuration>
-          <packagingExcludes>**/javax/**/*.*,**/client/**/*.class,**/*.symbolMap</packagingExcludes>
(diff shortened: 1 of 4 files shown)


manstis:
Can we have two <packagingExcludes> elements in the WARs' POM'S? There is already such an element containing a comma separated list a few lines above where you've added the new element.

psiroky:
Ouch, my bad. Not sure what's would the end result be. The maven-war-plugin was unfortunately not complaining. I will fix this asap.

psiroky:
Should be fixed now. I also split the excludes to one per line, so they are little bit more readable.

manstis:
+1 I'll merge on Tuesday (unless someone else fancies doing it earlier - public holiday here Monday)

manstis:
http://github.com/droolsjbpm/kie-wb-distributions/commit/5b8bf9a7e

PR #54, opened 2015-05-26
Remove CXF deps from KIE Drools WB as they are not needed

psiroky:
As discusses with @mswiderski last week, the CXF should not be needed for KIE Drools WB. This commit removes it from the distribution wars.


manstis:
http://github.com/droolsjbpm/kie-wb-distributions/commit/070c25f74

PR #55, opened 2015-05-28
[DROOLS-797] use different classifier for productized builds

psiroky:
* this way we can have both variants in same repo as Maven
   itself can differentiate between them using the classifier

@ryanzhang can you please take a look and let me know if this change somehow make you life harder? Basically all you need to change on your side are the artifact coordinates -- e.g. adding the new classifier.

### DIFF
--- a/kie-drools-wb/kie-drools-wb-distribution-wars/pom.xml
+++ b/kie-drools-wb/kie-drools-wb-distribution-wars/pom.xml
@@ -27,12 +27,6 @@
   </dependencyManagement>
 
   <dependencies>
-    <!-- Internal dependencies -->
-    <dependency>
-      <groupId>org.kie</groupId>
-      <artifactId>kie-drools-wb-webapp</artifactId>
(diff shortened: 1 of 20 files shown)


psiroky:
@manstis @porcelli please review, but don't merge yet. Let's wait for input from @ryanzhang.

manstis:
@psiroky Looks good to me.

porcelli:
@psiroky looks good to me too.

ryanzhang:
Ok,  it looks fine to me. 
(After conversation with Petr),  we think we might also want to do the same to dashbuilder.war which is producted in jbpm-dashboard.git.

psiroky:
Thanks Ryan. I will merge this and then look into the dashbuilder changes.

psiroky:
Merged into master as https://github.com/droolsjbpm/kie-wb-distributions/commit/fb10fee81ca58c31d1a1db76d704e7adeef56503

PR #57, opened 2015-06-01
Fixes GUVNOR-2189. Adds the META-INF dir of the 'cxf.impl' module to …

DuncanDoyle:
…the classpath.

See: https://issues.jboss.org/browse/GUVNOR-2189


psiroky:
+1. 

Note: the whole `cxf.impl` dependency is a thin ice as it is private module and we should not depend on it in ideal case. However, as long as we use itI don't see a reason why not to import the meta-inf.

We should probably do the same changes directly to the `kie-wb-webapp`. https://github.com/droolsjbpm/kie-wb-distributions/blob/master/kie-wb/kie-wb-webapp/src/main/webapp/WEB-INF/jboss-deployment-structure.xml#L59

DuncanDoyle:
Applied the same changes to the `kie-wb-webapp`.

psiroky:
Merged into master https://github.com/droolsjbpm/kie-wb-distributions/commit/af425745aad0cfd8b48c55f97fcb35f724ce9a4d. Thanks Duncan!

PR #58, opened 2015-06-02
GUVNOR-2187 : Real time verification and validation for decision tables

Rikkola:


manstis:
http://github.com/droolsjbpm/kie-wb-distributions/commit/34da96529

PR #59, opened 2015-06-02
BPMSPL-160: DRL text editor improvements

manstis:


Rikkola:
https://github.com/droolsjbpm/kie-wb-distributions/commit/d87492a42

Rikkola:
Looks like I can not close this one, but it is in.

PR #60, opened 2015-06-08
Selenium structure and basic test

ederign:
Depends on: https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/73

### DIFF
--- a/kie-wb-webapp-test/pom.xml
+++ b/kie-wb-webapp-test/pom.xml
@@ -17,7 +17,6 @@
     <version.wildfly>8.2.0.Final</version.wildfly>
   </properties>
 
-
   <dependencies>
     <dependency>
       <groupId>org.kie</groupId>
@@ -28,11 +27,6 @@
     <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
-      <groupId>org.apache.commons</groupId>
-      <artifactId>commons-exec</artifactId>
       <scope>test</scope>
     </dependency>
     <dependency>
--- a/kie-wb-webapp-test/pom.xml
+++ b/kie-wb-webapp-test/pom.xml
@@ -1,0 +1,164 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.kie</groupId>
+    <artifactId>kie-wb-distributions</artifactId>
+    <version>6.3.0-SNAPSHOT</version>
+  </parent>
+ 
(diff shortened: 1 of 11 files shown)


kie-wb/pom.xml
```diff
@@ -33,6 +33,7 @@
   <modules>
     <module>kie-wb-webapp</module>
     <module>kie-wb-distribution-wars</module>
+    <module>kie-wb-webapp-test</module>
```
porcelli: Can you setup this in a new maven profile? Also make sure that it also runs on fullProfile

psiroky: Maybe we could move this outside of the `kie-wb` submodule? I think it would be great to have the possibility to run the tests both on `kie-wb` and `kie-drools-wb` since lot of the functionality is in both.

psiroky:
It seems the `kie-wb` needs to be already deployed and running before executing the tests? If we want to keep this test module in "standard" or "full" build we need to make sure it won't break the build (which is the case now as the `kie-wb` needs to be deployed manually before the maven build).

kie-wb-webapp-test/pom.xml
```diff
@@ -0,0 +1,164 @@
...
+                <execution>
+                    <id>pre-integration-phase</id>
+                    <phase>pre-integration-test</phase>
+                    <goals>
+                        <goal>start</goal>
+                        <goal>deploy</goal>
+                    </goals>
+                </execution>
+            </executions>
+        </plugin>
```
psiroky: This looks like a perfect use case for cargo (codehaus-cargo.github.io). The lines `48` to `133` could be replaced by the following cargo config:
```xml
<plugin>
  <groupId>org.codehaus.cargo</groupId>
  <artifactId>cargo-maven2-plugin</artifactId>
  <configuration>
    <container>
      <containerId>wildfly8x</containerId>
      <type>installed</type>
      <zipUrlInstaller>
        <url>${wildfly8x.download.url}</url>
      </zipUrlInstaller>
    </container>
    <configuration>
      <properties>
        <cargo.jboss.configuration>standalone-full</cargo.jboss.configuration>
      </properties>
    </configuration>
    <deployables>
      <deployable>
        <groupId>org.kie</groupId>
        <artifactId>kie-wb-distribution-wars</artifactId>
        <classifier>wildfly8</classifier>
        <type>war</type>
      </deployable>
    </deployables>
  </configuration>
</plugin>
```
The advantage is of this approach is that we are not tied to any particular container. We can create different profiles for different containers and execute the tests easily on Tomcats, WildFly and EAP.

kie-wb-webapp-test/pom.xml
```diff
@@ -0,0 +1,164 @@
...
+            <excludes>
+              <exclude>none</exclude>
+            </excludes>
+            <includes>
+              <include>**/*Test.java</include>
+            </includes>
+          </configuration>
+        </execution>
+      </executions>
+    </plugin>
```
psiroky: We should generally use `maven-failsafe-plugin` for running integration tests. It is bound to `integration-test` phase by default.

porcelli:
https://github.com/droolsjbpm/kie-wb-distributions/commit/820ee58ceee39872437de794c17806747b10752f

PR #61, opened 2015-06-09
Remove 'x-ua-compatible' meta-data

manstis:
Remove 'x-ua-compatible' meta-data since (1) kie-(drools)-wb 6.3.0-SNAPSHOT doesn't run on IE11, (2) we've moved to GWT2.7 that supports IE11

See https://github.com/droolsjbpm/kie-wb-distributions/commit/bf6ab29d174b5c4d49b8d9512c75885190743d78

I also added meta-data tags to prevent caching of the host-page; that are inserted by the GWT2.7 Archetype.. so thought it prudent to use them too.


manstis:
Replaced by https://github.com/droolsjbpm/kie-wb-distributions/pull/68 (which only contains a single commit).

PR #71, opened 2015-06-22
BZ1230306: Guided Rule Editor: Dynamic enumerations do not work (documentation update)

manstis:


manstis:
http://github.com/droolsjbpm/kie-wb-distributions/commit/876cf25ed

PR #79, opened 2015-07-22
BZ1241414: Ability to set default locale of Business Central

manstis:
See https://bugzilla.redhat.com/show_bug.cgi?id=1241414


porcelli:
https://github.com/droolsjbpm/kie-wb-distributions/commit/c25d499e5

PR #81, opened 2015-07-26
readded kie deployment descriptor for all distributions kie-wb and ki…

mswiderski:
…e-drools-wb


manstis:
Thanks Maciejs, I'll try to look today.

manstis:
http://github.com/droolsjbpm/kie-wb-distributions/commit/c9a1862b9

Thanks Maciejs.

PR #84, opened 2015-07-30
Exclude org.uberfire.backend.server.LockClientNotifier that was causing system.git to be initialized ahead of kie-config-cli expecting it to be available.

manstis:
Hi Christian,

I noticed ```LockClientNotifier``` was setting up an instance of ```system.git``` (to hold the locks) ahead of when ```kie-config-cli``` was expecting it to be available. This lead to failure of the CLI tool to push the (```LockClientNotifier``` created) ```system.git``` to ```remote/origin``` before User's credentials had been configured on the FS (```LockClientNotifier``` registers ```system.git``` as "guest" whereas ```kie-config-cli``` needs ```system.git``` to be registered as a "real" user). 

Simple PR.. wordy explanation...

### DIFF
--- a/kie-config-cli/src/main/resources/META-INF/beans.xml
+++ b/kie-config-cli/src/main/resources/META-INF/beans.xml
@@ -8,6 +8,7 @@
   <weld:scan>
     <weld:exclude name="org.jbpm.kie.services.impl.form.FormProviderServiceImpl"/>
     <weld:exclude name="org.guvnor.structure.backend.repositories.git.GitRepositoryFactoryHelper"/>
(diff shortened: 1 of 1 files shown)


manstis:
This will need cherry-picking to the 6.3.x branch as well.

csadilek:
Thanks, Michael. This is the second time I broke something in the CLI without noticing ;(. Need to figure out how I can avoid that in the future. I will cherry-pick to 6.3.x.

manstis:
Hehe don't worry; kie-config-cli is the really, really and I mean really
poor relation *nobody* cares about!

Integration tests spring to mind.. but nobody takes real ownership of this
module yet alone want to write tests for it.... (yeah I include myself in
that too).

PR #85, opened 2015-08-06
JBPM-4722 - DataSet Definition auto detect datasource from persistenc…

mswiderski:
…e.xml

### DIFF
--- a/kie-wb/kie-wb-webapp/src/main/java/org/kie/workbench/backend/DashbuilderBootstrap.java
+++ b/kie-wb/kie-wb-webapp/src/main/java/org/kie/workbench/backend/DashbuilderBootstrap.java
@@ -25,11 +25,12 @@ import org.uberfire.commons.services.cdi.Startup;
 import javax.annotation.PostConstruct;
 import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
+import javax.persistence.EntityManagerFactory;
 
 @Startup
 @ApplicationScoped
 public class DashbuilderBootstrap {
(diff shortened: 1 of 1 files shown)


kie-wb/kie-wb-webapp/src/main/java/org/kie/workbench/backend/DashbuilderBootstrap.java
```diff
@@ -47,17 +48,31 @@
...
+    protected EntityManagerFactory emf;
+
     @PostConstruct
     protected void init() {
+        // figure out data source JNDI name
+        Object ds = emf.getProperties().get("hibernate.connection.datasource");
+        if (ds != null && ds instanceof javax.naming.Referenceable) {
+            try {
+                jbpmDatasource = ((javax.naming.Referenceable) ds).getReference().getClassName();
+            } catch (Exception e) {
```
Rikkola: What evil things will happen if this actually fails? Just thinking if we need to log or rethrow.

mswiderski: if it fails then it means it does not implement that interface and thus will fallback to this settings:
https://github.com/droolsjbpm/kie-wb-distributions/pull/85/files#diff-69c5ff9ccbae9219ce6f30a08dd82aecR33
either hardcoded or given as system property

Rikkola: Ok, thanks

PR #88, opened 2015-08-11
BPMSPL-209 - Provide alternative to DB based job executor

mswiderski:
enabling JMS based executor support that relies on this PR in jBPM repo: https://github.com/droolsjbpm/jbpm/pull/289


mswiderski:
merged into master and 6.3.x

PR #91, opened 2015-08-24
JBPM-4741 - Adding roles to to control access to the REST API

mrietveld:
See: https://issues.jboss.org/browse/JBPM-4741

Also see: 
- droolsjbpm/droolsjbpm-knowledge/pull/97
- droolsjbpm/guvnor/pull/153
- droolsjbpm/droolsjbpm-integration/pull/173

Please rebase this PR to preserve a linear history! :)

### DIFF
--- a/kie-wb/kie-wb-distribution-wars/src/main/assembly/tomcat7/WEB-INF/web.xml
+++ b/kie-wb/kie-wb-distribution-wars/src/main/assembly/tomcat7/WEB-INF/web.xml
@@ -115,6 +115,11 @@
 
   <!-- RESTEasy configurations -->
   <context-param>
+    <param-name>resteasy.role.based.security</param-name>
+    <param-value>true</param-value>
+  </context-param>
+
+  <context-param>
     <param-name>resteasy.scan</param-name>
     <param-value>true</param-value>
   </context-param>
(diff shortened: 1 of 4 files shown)


Rikkola:
And even though @mrietveld warned about linear history issues, I still managed to somehow not rebase this.

https://github.com/droolsjbpm/kie-wb-distributions/commit/83852bd15c0208eeae5c7e738ce21ae8dee477de

manstis:
What about kie-drools-wb?

mswiderski:
just out of curiosity - did this work on WebSphere? Asking mainly because of CXF was causing issues on WAS and thus this exclusion from container scanning (https://github.com/droolsjbpm/kie-wb-distributions/blob/master/kie-wb/kie-wb-distribution-wars/pom.xml#L394)  not sure it is relevant here...

mrietveld:
@manstis This is a kie-wb specific feature -- hmm.. uhh. Hmm! Well, I *thought* it was a kie-wb specific feature. I'll look into this, and if necessary, add it as well to the kie-drools-wb stuff.

mrietveld:
@mswiderski I wasn't able to test it on Websphere.. :/ I unfortunately had to do (educated by google) guesses about the was/cxf and wls/jersey configuration for this. 

I don't think that the container scanning should influence this: the `Ignore-Scanning-Archives` directive applies to all of the CXF archives, but not to the REST resource classes in `guvnor-rest-backend` and `kie-remote-services` -- and that's where the annotations are.

manstis:
@mrietveld kie-drools-wb contains the "screens" to manage execution servers.. so it's possible it'd need the same changes? I also see you're protecting ```ProjectResource``` which IIRC is the end-point for managing OUs, Repositories etc - which is definitely part of kie-drools-wb too.

mrietveld:
@manstis I've added this to kie-drools-wb as well. Thanks for the catch! 

master: https://github.com/droolsjbpm/kie-wb-distributions/commit/f669a89c0fc158a81762e718fb2f062664c75dd5 
6.3.x: https://github.com/droolsjbpm/kie-wb-distributions/commit/414335528c405fc2745ad9149fe03951089ce84b

PR #94, opened 2015-09-03
JBPM-4745 - Improve and update Remote API Documentation

mrietveld:
Please also cherry pick this to 6.3.x.

### DIFF
--- a/kie-drools-wb/kie-drools-wb-webapp/pom.xml
+++ b/kie-drools-wb/kie-drools-wb-webapp/pom.xml
@@ -194,6 +194,7 @@
       <artifactId>kie-internal</artifactId>
     </dependency>
 
+    <!-- identity provider -->
     <dependency>
       <groupId>org.kie</groupId>
       <artifactId>kie-identity-session-provider</artifactId>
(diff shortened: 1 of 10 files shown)


mrietveld:
@psiroky I'll do the content _generation_ stuff at a later date: at the moment, it just checks the content.

psiroky:
I am wondering if we could somehow reuse the code since it looks like both tests are doing almost the same thing.

Could we create a common module that the `kie-wb-webapp` and `kie-drools-wb-webapp` would depend on (test scope), which would contain that shared code? We then only need a simple test classes in both modules that would inherit the test code from the common module. WDYT?

mrietveld:
Yes, I was thinking the same thing. What about this? 

We create a "kie-wb-tests" module with
- kie-wb-smoke-tests
- kie-wb-webapp-tests

and then I can add the class to kie-wb-webapp-tests, and extend it in the kie-wb-*webapp modules?

psiroky:
`kie-wb-smoke-tests` actually depend on the created WARs, so they should be built after those. 

Could we move all the tests to a phase after the -webapps are built? It should be possible to extract the WAR using maven and look in there for the jsp file. I know it is not ideal, but I can't think of better solution (we could also divide the modules so that some are built before -webapps and some after, but that could cause confusion).

mrietveld:
I don't have a problem with that, but that's non-trivial to do. 

Would it be okay to merge this PR first and then do that second? Given that we're not talking about  a problem with the code itself, but about where/when the code should be and execute..

psiroky:
Sure, I will merge this asap and will think a bit about how to easily move the tests to a phase after the -webapps are built.

PR #99, opened 2015-09-09
bump permgen to 512m

qmx:
the previous setting gave me OOM: PermGen space.


psiroky:
There is a slight chance that we need the same bump for `kie-drools-wb-webapp` as well.

manstis:
*kie-wb-distributions*
(master) http://github.com/droolsjbpm/kie-wb-distributions/commit/0c5ed86e2

(6.3.x) http://github.com/droolsjbpm/kie-wb-distributions/commit/69a579e54

I also applied to ```kie-drools-wb-webapp``` to be safe, as suggested by @psiroky.

manstis:
Thanks @qmx

PR #102, opened 2015-09-17
[BZ-1207318] clean-up few deps to avoid duplicit classes with same FQN

psiroky:
* classes in the excluded jars are also provided
   by different jars. Having classes with same FQN
   in two different jars leads to hard to debug
   runtime issues as only one of those classes gets
   loaded (and there is no way to tell which one
   exactly)

 * there is few more duplicit classes that need to be looked at


psiroky:
Any thoughts on this one?

manstis:
@psiroky Looks OK to me.. I guess we need @tsurdilo to comment on the designer related changes.

psiroky:
These issues should be now fixed directly in designer. I am closing this one and will fix other possible issues directly in the designer.

PR #103, opened 2015-09-17
update readme about WebSphere setup to reflect an addition of new JMS…

rsynek:
… interfaces


rsynek:
Maciej, could you please take a look in case I forgot something?

mswiderski:
@rsynek I have copied same info to weblogic readme, hope you don't mind. thanks!

PR #105, opened 2015-09-24
[BZ-1207318] remove duplicated jaxrs jar in Tomcat distro

psiroky:
@mrietveld @mswiderski please take a look. We are bundling both `resteasy-jaxrs-api` and `jboss-jaxrs-api_1.1_spec` in the Tomcat WAR. One of them needs to be removed, I am just not 100 % sure which one. I picked the `jboss-jaxrs-api_1.1_spec`, but we may change that if needed.


psiroky:
This should be cherry-picked into 6.3.x as well (whatever solution we decide to go with).

mrietveld:
+1 it seems safer to keep the resteasy-jaxrs jar, given that it's version will match the rest of the resteasy code. 

Technically, there shouldn't be a difference -- although, it's possible we might move to a newer version of resteasy that implements the JAX-RS 2.0 API. In that case, this is the better choice, given that standards are (in my humble experience) not often as backwards compatible as we might want them to be..

PR #106, opened 2015-09-24
smoke-tests: add possibility to execute tests on prod binaries

psiroky:
* these changes enable execution of the tests on productized
   binaries (classifier ending with -redhat).
 * this should make it very easy to verify the productization
   process did not break anything (same tests need to pass before
   on community binaries and after on prod binaries)

* @mrietveld please take a quick a look in case I missed something

### DIFF
--- a/kie-wb-smoke-tests/README.md
+++ b/kie-wb-smoke-tests/README.md
@@ -25,3 +25,17 @@ different port.
 You can now run (and debug) the tests easily from IDE by just executing the `@Test` annotated
 methods.
 
+How to execute tests for different containers and apps
+======================================================
+By default, when you run `mvn clean install` _no_ tests will be executed. Tests are executed either when specifying `full`
+profile (by setting `-Dfull`) or when configuring explicit container and deployable profiles. The tests can be also executed
+on the productized binaries (by setting `-Dproductized`).
+
+Examples of different scenarios:
+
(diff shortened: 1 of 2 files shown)


mrietveld:
+1 on verifying productization process

@psiroky: what is important to me is that @mbiarnes should not be allowed to build a release without these tests passing! Even if it's just the tomcat tests (and not all platforms).

psiroky:
@mrietveld I ultimately agree. But it will need cooperation from others as well. When tests are failing (which means broken build), they should be fixed asap. We definitely should execute the whole matrix when doing releases (I am working on that as well).

Btw. I picked Tomcat because it the easiest one to use out-of-the-box. For EAP one would have to set the URL to download from and for WildFly, there is not productized WAR being built (it is not supported).

mrietveld:
okay, +1 to merge then. 

I would really like to make sure that these tests prevent all normal and full builds from succeeding -- will it be possible to have that in place by.. november?

psiroky:
I believe it is already possible (for master branch, but I am planning to backport this to 6.3.x asap). The failing tests will be reported and it just about not-ignoring them.

psiroky:
Note: the tests are executed only for `full` builds. I could not make this whole thing work otherwise. We are kind of abusing the use of Maven profiles here (but I don't really know better way).

PR #108, opened 2015-09-25
1266403 - fix for WebLogic kie server controller integration

sutaakar:
Fixes BZ 1266403.

May it cause some side effects?


mswiderski:
Don't think so as in many cases we do rely on the Jackson version we ship with the application (war file) so +1 to merge it.

PR #110, opened 2015-09-28
BZ1265652: Impossible to work with Business Central in IE10

manstis:
See https://bugzilla.redhat.com/show_bug.cgi?id=1265652


manstis:
The proposed fix does not solve issues - in fact it breaks IE11... rejecting.

PR #112, opened 2015-10-01
Impossible to work with Business Central in IE10

manstis:
See https://bugzilla.redhat.com/show_bug.cgi?id=1265652

This fix is only for *6.3.x* as the same fix does not work on master.


manstis:
See referenced BZ. Support for IE10 has been suspended pending feedback.

PR #114, opened 2015-10-05
Added note for Debian based systems

AmazingDreams:
I added an extra note for environment variables on Debian based systems as this caught me on the git/ssh authentication.


psiroky:
Merged into master, thanks @AmazingDreams!

PR #115, opened 2015-10-08
Selenium smoke test (login/logout)

jhrcek:
Here's the basic login/logout test (see LoginIntegrationTest). 

How to run it:

    cd kie-wb-distributions/kie-wb-smoke-tests
    mvn clean verify -Pwildfly8x,selenium,kie-wb

After having worked with Graphene/Drone for some time, once again I experienced the pain of writing UI test using plain selenium API. I think you'll experience it too if you try to expand the model & add some tests to do something more involved like navigating to authoring perspective, and opening an asset. Selenium API was not designed with testing javascript-heavy apps in mind. Although it IS possible the resulting code forces us to jump through hoops. But OK, let's try to stay with plain selenium for now and see where we can get.

### Next steps

1) Enable testing with InternetExplorer / Chrome. These require additional binary file to run [1] or [2] respectively. How do we automate obtaining these binaries in tests?

Important consideration: we want to avoid interacting with the DOM directly using JavascriptExecutor whenever possible. Because this enables us to achieve things in selenium tests that user interaction with the browser couldn't achieve (e.g. clicking an element that's hidden).

[1] https://sites.google.com/a/chromium.org/chromedriver/downloads
[2] http://selenium-release.storage.googleapis.com/index.html?path=2.48/

FYI @triceo


kie-wb-smoke-tests/pom.xml
```diff
@@ -349,6 +354,35 @@
...
+      <id>selenium</id>
+      <activation>
+        <property>
+          <name>selenium</name>
+        </property>
+      </activation>
+      <properties>
+        <deployable.groupId>org.kie</deployable.groupId>
+        <deployable.artifactId>kie-wb-distribution-wars</deployable.artifactId>
+        <deployable.context>kie-wb</deployable.context>
```
psiroky: These properties should not be needed here. It should be enough to activate the specific war profile (kie-wb, kie-drools-wb).

kie-wb-smoke-tests/src/test/java/org/kie/smoke/wb/selenium/model/LoginPage.java
```diff
@@ -0,0 +1,53 @@
...
+
+public class LoginPage extends PageObject {
+
+    @FindBy(name = "j_username")
+    private WebElement usernameInput;
+    @FindBy(name = "j_password")
+    private WebElement passwordInput;
+    @FindBy(className = "button")
+    private WebElement loginButton;
+    public final static String BASE_URL = "http://localhost:8080/kie-wb";
```
psiroky: The base URL is being passed to the tests via property (it is currently different for kie-wb and kie-drools-wb, we my need to change that in future), see https://github.com/droolsjbpm/kie-wb-distributions/blob/master/kie-wb-smoke-tests/src/test/java/org/kie/smoke/wb/AbstractWorkbenchIntegrationTest.java#L41 .

psiroky:
Thanks @jhrcek! Looks really good, there are just few configuration related things I mentioned in the comments. Could you please fix them?

I will also try to run the test locally.

jhrcek:
@psiroky Thanks for reviewing, I updated what you mentioned. Now it should be possible to run the tests using 

    mvn clean verify -Pwildfly8x,selenium,kie-wb

psiroky:
I added selenium-bom into the parent hierarchy (https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/commit/8cf84176ffa6dc422cf68cd4e928fadda2a23c0b), so that should fix the issue with enforcer and transitive deps (after adding `selenium-api, selenium-firefox-driver, selenium-support, guava`). @jhrcek could you please try now and move the supporting classes into src/main/ as you originally mentioned? Let me know in case you would encounter some additional issues.

jhrcek:
Supporting classes moved to `src/main/java` and PR comments updated. I appreciate your help @psiroky

triceo:
@psiroky  Every now and then, we need to update Selenium. (As new browsers are introduced, and legacy browsers removed.) How much flexibility is there for us to do this?

psiroky:
@triceo I will need to figure out if we need to move the dependencies to IP BOM. Since they are test-scope, I think it should not be necessary. If we keep them in our kie-platform-bom then it is simply a matter of creating PR with the change and we will accept it after making sure nothing breaked.

In case the dependencies would be pushed to IP BOM, it would be slightly difficult as there are multiple parties which need to ack the changes. But it should not take more then couple of days to get the changes even there.

psiroky:
+1. Thanks @jhrcek!

csadilek:
w00t! Thanks, @jhrcek and @psiroky.

master:
a4f8df62887327a3e367821e18ea0461d902eff5
0a6d49a0048a9338e3fcdeb39d3e19d622945fe5

PR #118, opened 2015-10-13
BZ-1270872: E(J)WS,WLS: SessionListener for releasing locks not invoked

csadilek:


psiroky:
Is there a chance we would need the same change for the WebSphere assemblies as well?

csadilek:
@psiroky there is, but it should work as is in theory (based on my research and tests from a while ago). We'll need QE to verify. Let's merge and (re-)test?

csadilek:
After more tests and enabling more detailed logging, it turns out that the SessionListener is in fact invoked correctly on Tomcat and E(J)WS, but the injected IOService is null on these containers. We need to investigate. Closing this as not required.

PR #119, opened 2015-10-16
Drop IE9 and IE10 GWT permutations

psiroky:
* we only need to support IE 10+

### DIFF
--- a/kie-drools-wb/kie-drools-wb-webapp/src/main/resources/org/kie/workbench/drools/KIEDroolsWebapp.gwt.xml
+++ b/kie-drools-wb/kie-drools-wb-webapp/src/main/resources/org/kie/workbench/drools/KIEDroolsWebapp.gwt.xml
@@ -106,7 +106,8 @@
   <extend-property name="locale" values="zh_CN"/>
   <extend-property name="locale" values="de"/>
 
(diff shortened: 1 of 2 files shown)


manstis:
I believe IE10 has also been dropped (at least in product).

@krisv @etirelli Can either of you please advise whether we can drop IE10 in community too?

psiroky:
Yes, I believe we could actually do that. In case we need to support IE 10 again in future, we could always add it back.

etirelli:
@manstis @psiroky for community we can drop IE10, but for product unfortunately I don't think a decision was made yet. I am trying to confirm this and will let you know.

manstis:
OK,  that's not the message Kris was giving me earlier this week!

etirelli:
@manstis @psiroky Prakash just confirmed we can drop it for now. So go for it. :)

psiroky:
I've updated the PR accordingly.

manstis:
+1 to merge.

csadilek:
Thanks!

master: 402487a4fb649882720c82e97114302b20b764c5
6.3.x: 0fb0d806e2ae7e6700d907c2c33d4e6bb4992c55

PR #120, opened 2015-10-16
[DROOLS-955] remove kie-drools-wb-distribution module

psiroky:
The resulting zip is too big (almost 900MiB). We will list individual WARs on the downloads page instead. See https://github.com/droolsjbpm/drools-website/pull/10 for Drools Website updates.


manstis:
+1 from me!

Rikkola:
+1 
I have downloaded the zip few times and even when it was smaller it took time and I am probably one of the few people that ever needed wars for different containers.

csadilek:
Thanks!

master: 32caa9a6e6d9deb669cb6abc0bf1f4c35c56c576

PR #123, opened 2015-10-21
BZ:1273362 - Databases: DataSetLookupException in server.log (Master)

nmirasch:
Reviewed the query to show domain variables. Tested in MySql, Oracle, PostgreSQL and H2.


Salaboy:
+1

csadilek:
@Salaboy @nmirasch This might be one of the edge cases where we don't require a unit test as part of the fix. Only an integration-test would have caught that problem and can guard against regressions in the future. Would you mind creating a JIRA for writing an integration test? /cc @jhrcek.

csadilek:
master: e01ee86dd1e12aa1f736d6c36e65d31b21148017

jhrcek:
@Salaboy @nmirasch  @csadilek +1 to the integration tests idea. I'm afraid this functionality (SQL query-based datasets) will be source of more bugs. The 3-5 UI tests we have cover the filter creation functionality cover the functionality very sporadically, yet they discovered quite a few issues when executed against all the supported databases. I'm afraid more bugs are lurking there.

Salaboy:
@csadilek @jhrcek totally agree, we are planning to get rid of that SQL nightmare, unfortunately that's what we have right now. Regarding the tests, we are also planning to move these screens to MVP and add tests in the future

PR #124, opened 2015-10-23
InternetExplorer setup in selenium smokes

jhrcek:
THIS PR IS MERGE READY

Setup for using InternetExplorerDriver in selenium smoke tests.
I configured existing driver-binary-downloader-maven-plugin to download, cache and unzip IEDriverServer.exe binary required to run InternetExplorerDriver. This plugin also sets webdriver.ie.driver maven property, which is made available to tests via failsafe plugin configuration.

Example usage:
    mvn clean install -Pwildfly8x,selenium,kie-wb # use Firefox by default
    mvn clean install -Pwildfly8x,selenium,kie-wb,ie # ie profile enables tests with InternetExplorer (windows only)

I tested this on Windows 8 with IE 11 and it's working fine. During the testing I encountered one long-known issue with 64bit IEDriverServer [1]. This issue causes typing into input elements to be very slow (1 character per 5 secons). I applied the know workaround, which is to use 32bit IEDriverServer binary on both 32 and 64bit machines.

[1] https://code.google.com/p/selenium/issues/detail?id=5116

### DIFF
--- a/kie-wb-smoke-tests/pom.xml
+++ b/kie-wb-smoke-tests/pom.xml
@@ -94,6 +94,10 @@
       <artifactId>selenium-firefox-driver</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.seleniumhq.selenium</groupId>
+      <artifactId>selenium-ie-driver</artifactId>
+    </dependency>
+    <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
     </dependency>
@@ -367,8 +371,10 @@
     </profile>
 
     <profile>
-      <!-- Enable selenium-based smoke tests to run. E.g.
-           mvn clean install -Pwildfly8x,selenium,kie-wb  -->
+      <!-- Enable selenium UI smoke tests to be run with Firefox. 
+      Example usage: mvn clean install -Pwildfly8x,kie-wb,selenium
+           
+      To use Internet Explorer see profile ie -->
       <id>selenium</id>
       <activation>
         <property>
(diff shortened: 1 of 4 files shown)


kie-wb-smoke-tests/pom.xml
```diff
@@ -385,6 +391,27 @@
                 <groups>org.kie.smoke.wb.category.KieWbSeleniumSmoke</groups>
               </configuration>
             </plugin>
+            <plugin> <!-- Download & unzip IEDriverServer.exe - a binary required to run InternetExplorerDriver -->
```
psiroky: To make the `<execution>` work it needs to be moved directly to `<plugins>` section of the POM (instead of `<pluginManagement>`)

kie-wb-smoke-tests/src/main/java/org/kie/smoke/wb/selenium/model/WebDriverFactory.java
```diff
@@ -0,0 +1,51 @@
...
+import org.openqa.selenium.firefox.FirefoxDriver;
+import org.openqa.selenium.ie.InternetExplorerDriver;
+
+/**
+ * Class responsible for instantiating WebDriver instance.
+ */
+public class WebDriverFactory {
+
+    // IEDriverServer binary downloaded by antrun plugin in pom.xml
+    private static final File IE_DRIVER_BINARY = new File(System.getProperty("java.io.tmpdir"), "IEDriverServer.exe");
```
psiroky: Although the filename won't likely change, I agree that it would be better to pass property to the test via failsafe plugin configuration.

kie-wb-smoke-tests/pom.xml
```diff
@@ -385,6 +391,27 @@
...
+                  <goals>
+                    <goal>run</goal>
+                  </goals>
+                  <configuration>
+                    <target>
+                      <get src="http://selenium-release.storage.googleapis.com/2.48/IEDriverServer_x64_2.48.0.zip"
+                           dest="${java.io.tmpdir}"
+                           skipexisting="true"/>
+                      <unzip src="${java.io.tmpdir}/IEDriverServer_x64_2.48.0.zip"
+                             dest="${java.io.tmpdir}"
```
psiroky: I think that we should unzip this into target/ so that we can rely on "clean" builds. When we depend on some content in /tmp maven will not clean that up and we may end up with some weird bugs related to old version of the IE driver in the /tmp. In ideal case, we should download the zip file only once (and cache it) and unzip it with every clean build. This is what download-maven-plugin should be able to do.

kie-wb-smoke-tests/pom.xml
```diff
@@ -385,6 +391,27 @@
...
+              <artifactId>maven-antrun-plugin</artifactId>
+              <executions>
+                <execution>
+                  <phase>pre-integration-test</phase>
+                  <goals>
+                    <goal>run</goal>
+                  </goals>
+                  <configuration>
+                    <target>
+                      <get src="http://selenium-release.storage.googleapis.com/2.48/IEDriverServer_x64_2.48.0.zip"
```
psiroky: +1. We need to figure out how to make this work as simple property replacement would work here (2.48 vs 2.48.0).

csadilek:
Thanks, great stuff!!

master: b93c017f1117e60e59ffa8635633d3a211c3d60f

psiroky:
I will look into automating this in our CI.

PR #128, opened 2015-11-03
enabled JMS based executor by using standalone-full.xml profile, copy…

mswiderski:
… webapp sources into kie-wb.war folder so wildfly recognizes it as web app

this is more to allow developers working on asset mgmt to use hosted mode. Not sure this is the best way to address the problem but at least it does work so might be worth local merge to ease development.

In general what has been done here:
- use standalone-full.xml profile to allow use of JMS resources in kie-wb 
- copy entire web app into folder that must be suffixed with .war otherwise wildfly does not recognize it as web app and skips reading some descriptors such as ejb-jad.xml see:https://github.com/wildfly/wildfly/blob/master/ejb3/src/main/java/org/jboss/as/ejb3/deployment/processors/EjbJarParsingDeploymentUnitProcessor.java#L213
- change in extraJvmArgs is just to put them in single line as otherwise hosted mode didn't start for me... no idea why....

### DIFF
--- a/kie-wb/kie-wb-webapp/pom.xml
+++ b/kie-wb/kie-wb-webapp/pom.xml
@@ -39,6 +39,7 @@
     <errai.jboss.home>${project.build.directory}/wildfly-${as.version}</errai.jboss.home>
     <gwt.compiler.localWorkers>1</gwt.compiler.localWorkers>
     <gwt.memory.settings>-Xmx4g -Xms1g -XX:MaxPermSize=512m -XX:PermSize=128m -Xss1M</gwt.memory.settings>
+    <webapp.dir>src/main/webapp</webapp.dir>
   </properties>
 
   <dependencies>
@@ -1493,9 +1494,7 @@
           <configuration>
             <deploy>${project.build.directory}/gwt-symbols-deploy</deploy>
             <localWorkers>${gwt.compiler.localWorkers}</localWorkers>
-            <extraJvmArgs>${gwt.memory.settings} -XX:CompileThreshold=7000 -Derrai.jboss.home=${errai.jboss.home} 
-              -Derrai.marshalling.server.classOutput=${project.build.outputDirectory} -Dorg.kie.demo=true 
-              -Dorg.kie.clean.onstartup=true -Dorg.uberfire.nio.git.ssh.enabled=false -Djava.util.prefs.syncInterval=2000000
+            <extraJvmArgs>${gwt.memory.settings} -XX:CompileThreshold=7000 -Derrai.marshalling.server.classOutput=${project.build.outputDirectory} -Dorg.kie.demo=true -Dorg.kie.clean.onstartup=true -Dorg.uberfire.nio.git.ssh.enabled=false -Djava.util.prefs.syncInterval=2000000 -Derrai.jboss.home=${errai.jboss.home}
             </extraJvmArgs>
             <draftCompile>true</draftCompile>
             <module>org.kie.workbench.FastCompiledKIEWebapp</module>
@@ -1504,7 +1503,7 @@
             <server>org.jboss.errai.cdi.server.gwt.EmbeddedWildFlyLauncher</server>
             <disableCastChecking>true</disableCastChecking>
             <runTarget>kie-wb.html</runTarget>
-            <hostedWebapp>src/main/webapp</hostedWebapp>
+            <hostedWebapp>${webapp.dir}</hostedWebapp>
             <!-- drools-compiler has dependency on org.eclipse.jdt.core.compiler:ecj:jar:3.5.1:compile, see http://code.google.com/p/google-web-toolkit/issues/detail?id=4479 -->
             <gwtSdkFirstInClasspath>true</gwtSdkFirstInClasspath>
             <compileSourcesArtifacts>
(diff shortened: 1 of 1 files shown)


Salaboy:
+1

mswiderski:
and this is the reason why we need it in first place: https://issues.jboss.org/browse/WFLY-4198

kie-wb/kie-wb-webapp/pom.xml
```diff
@@ -1818,6 +1816,49 @@
           </execution>
         </executions>
       </plugin>
+      <plugin>
+        <artifactId>maven-antrun-plugin</artifactId>
+        <version>1.4</version>
```
psiroky: The version is already defined in `kie-parent-metadata` (see https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/blob/master/pom.xml#L368), so should be reused.

kie-wb/kie-wb-webapp/pom.xml
```diff
@@ -1818,6 +1816,49 @@
...
+            </configuration>
+            <goals>
+              <goal>run</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
+      <plugin>
+        <artifactId>maven-antrun-plugin</artifactId>
+        <version>1.4</version>
```
psiroky: Same as above, should be possible to delete the version tag and use the one coming from `kie-parent-metadata`.

manstis:
The only difference between running kie-wb in hosted mode vs deploying the
Wildfly WAR to a standalone Wildfly instance is the inclusion of slf4j and
logback JARs in WEB-INF/lib in hosted mode. I am experimenting with
removing all logging implementations when running in hosted mode.....

kie-wb/kie-wb-webapp/pom.xml
```diff
@@ -1504,7 +1502,7 @@
             <server>org.jboss.errai.cdi.server.gwt.EmbeddedWildFlyLauncher</server>
             <disableCastChecking>true</disableCastChecking>
             <runTarget>kie-wb.html</runTarget>
-            <hostedWebapp>src/main/webapp</hostedWebapp>
+            <hostedWebapp>${project.build.directory}/kie-wb.war</hostedWebapp>
```
psiroky: Would it be now possible to fix the output directory directly here https://github.com/mswiderski/kie-wb-distributions/blob/hosted-mode-executor-fix/kie-wb/kie-wb-webapp/pom.xml#L1475 instead of copying stuff around (e.g. those antrun executions)?

psiroky: I believe that would also be better practice as generated stuff should not go into src/main, but rather target/something.

mswiderski: src/main/webapp is part of sources and thus stored in git and that's the reason why I did copy it to target instead. Note that there are bunch of staff copied there during the build. Anyway not sure why src/main/webapp is used for this purpose so can't say for sure...

psiroky: We should be able to configure the maven in way that all resources from src/main/webapp are copied into target/something. So basically the result would be the same, but there would be less configuration in the pom.

manstis:
@mswiderski, @Salaboy, @porcelli, @csadilek 

IMO this "smells bad".. I know it's a (partial) work-around for hosted-mode but it just doesn't feel the right thing to do. We still have a bunch of other Async EJB's that will fail to run on WildFly <10.0.0.CR4 in hosted mode. Personally, I'd prefer a global solution that'll fix all Async EJB's. 

What's so different about hosted mode? 

* The *real* WAR doesn't include any logging implementation JARs whereas hosted mode does
* Hosted mode is launched with EmbeddedWildFlyLauncher which sets up a Log Manager (IDK if running "regular" WF does the same - I tested removing the Log Manager setup and WF embedded failed to start...)

Should however we decide this really is the only solution (to jBPM's ExecutorService at least) then I think it'll need applying equally to kie-drools-wb, as that has Asset Management too and hence by implication uses jBPM's Asyc EJB ExecutorService.

csadilek:
An alternative I see is to enhance EmbeddedWildFlyLauncher to do all of the above (use standalone-full.xml instead and use a directory that ends in .war for deploying) based on a configuration. It already copies others (i.e. application-users and application-roles.properties). If you think that's preferable, please create two JIRAs in Errai. I guess Michael's concern is not really addressed by that as we'd just move the workaround out of the way.

mswiderski:
Main intention of this PR was to unblock people working on asset mgmt issues. 
Personally I am not completely happy with the solution as it requires lots of resource manipulation. Thus I like Christian idea to enhance embedded launcher instead. 

Mike, when it comes to executor and why we could workaround it was recent enhancement to jbpm executor that allows use of jms for better performance. Though jms based was not added to kie drools wb to keep it simple as that would require jms queue configuration to be able to deploy app. Not a big issue for eap/wildfly though more advanced for other containers. 

So I am completely ok with dropping this PR from the list

Dnia 03.11.2015 o godz. 21:41 Christian Sadilek <notifications@github.com> napisał(a):

manstis:
Ok, thanks for the explanation Maciejs! Whilst I don't particularly like
this PR I also feel the pain of others' of not being able to run code in
Hosted Mode. So if this really is blocking salaboy then we possibly have no
short term alternative.

mswiderski:
alternatively we can move this to a profile as we consider it anyway just a workaround. Will be easier to remove entire provide instead of individual lines. Wdyt?

manstis:
+1 for a profile for the partial workaround.

mswiderski:
profile created instead

@csadilek jiras for Errai created and mentioned in profile that once they are resolved profile can be removed:
https://issues.jboss.org/browse/ERRAI-887
https://issues.jboss.org/browse/ERRAI-888

now if you need executor to be available in hosted mode you can run get with:
```
mvn clean gwt:debug -PhostedMode
```

csadilek:
Thanks!

6.3.x: 796dade30861d1b2c7649bc33369a99f340211e5

PR #130, opened 2015-11-06
BZ-1275796 - Fixed banner and navigation menu styling to match Patter…

cristianonicolai:
…nFly guideline for the product


manstis:
Where is banner.html used?

cristianonicolai:
banner.html is used to override the same file from the community version. In this case the difference is the color scheme, using red and grey instead of blue and black.

manstis:
+1 to merge.

csadilek:
Thanks!

master: 803eb5e18d78e6f57911fd2cf237d976ae173b78

PR #132, opened 2015-11-10
JBPM-4821 - Use perspective menu in order to keep selection state

cristianonicolai:


cristianonicolai:
Depends on https://github.com/uberfire/uberfire/pull/212 for initial selection to work.

csadilek:
This is a +1 but I will wait for other more familiar to take a look i.e. @manstis?

manstis:
To be honest I don't know the cause/need of this PR but accepting a lack of knowledge the commit itself seems ok.

cristianonicolai:
@csadilek @manstis, this change is only that the navigation bar can keep the current selection. That means every time we navigate to a different perspective, it try to match a menu item to set as active. Technically, just a different color in the nav bar. To keep track of this selection, I created a new type of menu item called MenuItemPerspective, which in this PR I had to update all menus to this new type. Hope that helps :)

manstis:
@cristianonicolai Thanks for the explanation.

+1 to merge.

csadilek:
Thanks!

master: 13db23a6eb452c325370c687dbfa4a27e36a7d76

PR #135, opened 2015-11-12
JBPM-4812 - Allow to query by task variables - added new entity to pe…

mswiderski:
…rsistence.xml


mswiderski:
depends on https://github.com/droolsjbpm/jbpm/pull/338

PR #146, opened 2015-11-23
(6.3.x) BZ1192831: User with no privileges for Repository can view and modify assets in that Repository. Ensure the same instance of RuntimeAuthorizationManager is used.

manstis:
See https://bugzilla.redhat.com/show_bug.cgi?id=1192831

This is the final piece of the jigsaw.. multiple instances of ```RuntimeAuthorizationManager``` were being created which was leading to authorization behaving unpredictably.


manstis:
@csadilek Tests added (this BZ is on the "blocker" list for Product 6.2.x)

csadilek:
Thanks.

6.3.x: 145f992e6c419a62de79c58ad9a14ee4c89c9bc1

PR #147, opened 2015-11-23
BZ1192831: User with no privileges for Repository can view and modify assets in that Repository. Ensure the same instance of RuntimeAuthorizationManager is used.

manstis:
See https://bugzilla.redhat.com/show_bug.cgi?id=1192831

This is the final piece of the jigsaw.. multiple instances of RuntimeAuthorizationManager were being created which was leading to authorization behaving unpredictably.

(cherry picked from commit 47e5054a1fa55dbb058cbcdd671475335ba3d095)

### DIFF
--- a/kie-drools-wb/kie-drools-wb-webapp/src/main/java/org/kie/workbench/drools/backend/server/ApplicationScopedProducer.java
+++ b/kie-drools-wb/kie-drools-wb-webapp/src/main/java/org/kie/workbench/drools/backend/server/ApplicationScopedProducer.java
@@ -47,12 +47,13 @@ import org.uberfire.security.impl.authz.RuntimeAuthorizationManager;
 @ApplicationScoped
 public class ApplicationScopedProducer {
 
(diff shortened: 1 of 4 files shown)


csadilek:
@manstis should we add a test for this? Basically making sure we always get the same instance of AuthManager...something like:

ApplicationScopedProducer asp = new ApplicationScopedProducer();
AuthorizationManager am1 = asp. getAuthManager();
AuthorizationManager am2 = asp. getAuthManager();
assertSame(am1, am2);

Since this was hard to track down it would guard against regressions in case someone refactors this to lazily initialize the auth manager.

manstis:
@csadilek Good idea (yes it was difficult to track down); I'll write the test tomorrow and update the PRs. Thanks

manstis:
@csadilek Tests added.

csadilek:
Thanks!

master: a2ab48e697611b4ed190be8ae40336f13f3f7d93

PR #149, opened 2015-11-23
6.3.x - BZ1284004 and BZ1283537

nmirasch:
Changed SQL queries that define the dataset

### DIFF
--- a/kie-wb/kie-wb-webapp/src/main/java/org/kie/workbench/backend/DashbuilderBootstrap.java
+++ b/kie-wb/kie-wb-webapp/src/main/java/org/kie/workbench/backend/DashbuilderBootstrap.java
@@ -57,11 +57,6 @@ public class DashbuilderBootstrap {
     @Inject
     protected DataSetDefRegistry dataSetDefRegistry;
 
-
-    public void setDataSetDefRegistry(DataSetDefRegistry dataSetDefRegistry){
-        this.dataSetDefRegistry = dataSetDefRegistry;
-    }
-
     @PostConstruct
     protected void init() {
         // figure out data source JNDI name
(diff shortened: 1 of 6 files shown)


kie-wb/kie-wb-webapp/src/main/java/org/kie/workbench/backend/DashbuilderBootstrap.java
```diff
@@ -57,6 +57,11 @@
     @Inject
     protected DataSetDefRegistry dataSetDefRegistry;
 
+
+    public void setDataSetDefRegistry(DataSetDefRegistry dataSetDefRegistry){
```
cristianonicolai: If you use @InjectMocks in the test, Mockito would that care of injecting an instance of DataSetDefRegistry. That way this method wouldn' t be necessary.

kie-wb/kie-wb-webapp/src/test/java/org/kie/workbench/backend/DashbuilderBootstrapTest.java
```diff
@@ -0,0 +1,52 @@
...
+import org.dashbuilder.dataset.def.DataSetDefRegistry;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+
+import static org.mockito.Matchers.any;
+import static org.mockito.Mockito.*;
+
+@RunWith(GwtMockitoTestRunner.class)
```
cristianonicolai: Because there is no GWT code under test here, you could simply use MockitoJUnitRunner instead. That also makes adding both dependencies gwtmockito and uberfire-testing-utils unnecessary at this stage.

nmirasch: ok, let me review that

csadilek:
+1

master: bd437e6b9fcb92430bb24a46e644dc761eaef278 and bfcb6c29b36684a763cc2e679c8e7acc42a31c93

PR #153, opened 2015-11-25
[BZ-1020902] exclude source files from WARs

psiroky:
Source files are useless in the WAR, so no need to have them there. 

Adjusted a bit the other exclude patterns to make it clearer what they are targeting.


manstis:
+1 to merge.

csadilek:
Thanks.

master: 62e79f150c7b7bcfe7433005a5dfba0a1b9414a3

PR #157, opened 2015-11-30
Do not disable other smoke tests when running Selenium tests

psiroky:
* before the change, "selenium" profile would overwite
   the failsafe config causing execution of only the
   seleinum based tests (others would be ignored). This
   change adds new failsafe execution that runs the selenium
   tests. It will be executed after the other smoke tests are
   ran.
 * by default Selenium tests are still disabled (no change here).
   One needs to enable 'selenium' profile to run them.

@jhrcek please take a look. The change effectively means that using `mvn clean install -Pkie-wb,tomcat7x,selenium` will run other smoke tests (REST, etc) as well, not just the selenium ones like so far. This should simplify execution of the tests in CI and also locally. I just want to make sure I did not overlook something.

### DIFF
--- a/kie-wb-smoke-tests/pom.xml
+++ b/kie-wb-smoke-tests/pom.xml
@@ -371,10 +371,12 @@
     </profile>
 
     <profile>
-      <!-- Enable selenium UI smoke tests to be run with Firefox. 
-      Example usage: mvn clean install -Pwildfly8x,kie-wb,selenium
-           
-      To use Internet Explorer see profile ie -->
+      <!-- Enable selenium UI smoke tests to be run with Firefox.
+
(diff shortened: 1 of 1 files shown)


jhrcek:
I agree with the change if we somehow still retain the option to run the selenium test separately for debugging purposes (like with -Dtest=LoginLogoutTest). Can you please make sure something like that will be possible?

psiroky:
Good point. I will make sure that is still possible.

psiroky:
I just verified that it is still possible to run just single test using `-Dit.test` (can't use `-Dtest` as that will be picked up by surefire will fail it won't find the IntegrationTest class). I've also added an example of that in the pom.

jhrcek:
Thanks, +1 to merge.

PR #158, opened 2015-12-02
 JBPM-4857 - splitting up kie-wb-smoke-tests so that both cargo and arquillian tests are possible

mrietveld:
@psiroky Could you look at this?


mrietveld:
@psiroky Updated PR to have less code. 

Some of the duplicated code will also be deleted in the follow-up PR.

mrietveld:
jenkins, please retest

mrietveld:
@psiroky The *only* reason to have arquillian tests is that it means that it's a lot easier for me (and others?) to develop tests for the cargo module. 

However, running "mvn clean install -Pkie-wb,eap64x", for example, will result in test failures because arquillian is not well suited to running with maven, especially when there are multiple test classes. 

Does this make sense to you? Let me know if you have concerns.

psiroky:
@mrietveld I have been looking at the changes and I am still worried about the duplication between `cargo` and `arquillian` modules. For example all the container configuration files are currently in both modules. This will make it very easy for people to just forgot to update both places at the same time.

So I have been thinking about how to share as much stuff as possible between those and I came up with two possibilities:

  1. Leaving the `kie-wb-smoke-tests` module as is and create new one `kie-wb-smoke-tests-arq` which would depend on the `kie-wb-smoke-tests`. The ARQ tests would just instantiate test classes from the other module and run the test methods inside their own @Test annotated methods. This is far from ideal, but it would keep the cargo-based untouched and easy to use.

  2. Having the same structure as you are proposing (base, cargo, arq), but leave the all the shared files in `base` and use them in those two other modules. This would likely mean that we would go outside the module dirs (e.g. ../kie-wb-smoke-tests/src/test/resources) to copy the config files, but I believe that should be fine in this case, because we would only use static resources, not those generated by the build itself (we would not copy stuff from ../kie-wb-smoke-tests/target).

WDYT? Would any of these alternatives work for you?

mrietveld:
@psiroky Please submit your own PR to do this, either option is fine.

psiroky:
OK, I will submit new PR once I am done.

PR #165, opened 2015-12-14
Add possiblity to run smoke tests via Arqullian

psiroky:
* Cargo tests still working as before, but are in different module
 * ARQ tests need more work
 * re-organized resources for smoke tests to enable better sharing

@mrietveld please take a look.

Note: I will take care of merging this once we reviewed, because it requires additional fixes to some Jenkins jobs (I changed names of the profiles to sync them with WAR classifiers).


mrietveld:
+1 please merge if build completes successfully or with otherwise unrelated failures.

PR #166, opened 2015-12-15
Don't smoke test some perspectives when running agains kie-drools-wb

jhrcek:
Exclude some perspectives from LoadAllPerspectivesIntegrationTest (selenium smoke which tries to navigate to all perspectives and checks that they're loaded). The exclusion is done based on system property app.name which is set to kie-wb by default and overriden to kie-drools-wb when running smokes with kie-drools-wb maven profile.


psiroky:
+1, merged into master.

PR #171, opened 2016-01-12
Fix smoke tests config for Tomcat 8

psiroky:
* without the fix it was not possible to
   test prod (-redhat) WARs on Tomcat 8

Same fix was already merged to master. There is no approved BZ for this, but since these are just test fixes, I believe it is OK to be merged into 6.3.x.


manstis:
+1 to merge.

csadilek:
6.3.x: 0f78406a962a5c9807736c92ce8d6d5cf58f10b1

PR #172, opened 2016-01-13
JBPM-4901: UberFire User System Management integration into the Workbenches

romartin:
NOTE: This PR depends on the code updates currently in https://github.com/uberfire/uberfire-extensions/pull/173.

Thanks!


csadilek:
@manstis @psiroky can you guys take a look at this as well, please?

@romartin as discussed, this is missing the keycloak option/dependencies. How do you plan to support that? Can you summarize the configuration options here so that others know as well. Thanks.

romartin:
Please, @csadilek , @psiroky , @manstis @krisv @etirelli can you confirm the idea:

As you know, we have security management providers for KeyCloak, EAP, Wdilfly and Tomcat. So in both KIE-WB and drools-wb, that's my opinion:

- The showcase - > only comes with the wf/eap provider dependency, as it runs using the errai embedded wildfly
- The tomcat distro -> comes only with the tomcat dep

romartin:
FYI. Adding keycloak implies this dep:

<dependency>
        <groupId>org.uberfire</groupId>
        <artifactId>uberfire-security-management-keycloak</artifactId>
 </dependency>

That transitively.... needs the resteasy client artifacts and the custom keycloak rest client artifacts, so:


    <dependency>
      <groupId>org.keycloak</groupId>
      <artifactId>keycloak-core</artifactId>
    </dependency>

    <dependency>
      <groupId>org.keycloak</groupId>
      <artifactId>keycloak-model-api</artifactId>
    </dependency>

    <dependency>
      <groupId>org.keycloak</groupId>
      <artifactId>keycloak-adapter-core</artifactId>
    </dependency>

    <dependency>
      <groupId>org.keycloak</groupId>
      <artifactId>keycloak-account-api</artifactId>
    </dependency>

    <dependency>
      <groupId>org.keycloak</groupId>
      <artifactId>keycloak-events-api</artifactId>
    </dependency>

    <dependency>
      <groupId>org.keycloak</groupId>
      <artifactId>keycloak-admin-client</artifactId>
    </dependency>

    
    <dependency>
      <groupId>org.jboss.resteasy</groupId>
      <artifactId>jaxrs-api</artifactId>
      <optional>true</optional>
    </dependency>
    <dependency>
      <groupId>org.jboss.resteasy</groupId>
      <artifactId>resteasy-client</artifactId>
      <optional>true</optional> 
    </dependency>

manstis:
@romartin @csadilek My comments are:-

* It appears that "User and Group Management" is being added to the "Home" menu.. surely adding them to "Administratiion" would seem a more logical placement?

* "Out of the Box" configuration for Tomcat and EAP/Wildfly seems to be supported. @romartin Could you please clarify the position for WAS and WebLogic? Do these distributions have the feature disabled "out of the box" or do they work with the ```uberfire-security-management-wildfly-*.jar``` JAR (as it doesn't appear to be removed with their respective assembies).

* The configuration needs detailed documentation. I've not seen a PR for this yet?

psiroky:
The build failed because the branch names do not match (the UF PR branch is called `users-management-fix` and branch for this PR is called `user-system-management`). Since the first PR has been already merged, it should be enough to just rerun the build (just triggered). It would be great to see the CI build pass with these changes.

About the PR changes:
 *  like @manstis mentioned we need to make clear if we support WebSphere and/or WebLogic. 

 * looking at the keycloak client transitive deps, I don't think we can simply bundle them inside the WAR (be it by default or by patch). The `resteasy-client` depends on `resteasy-jaxrs` which is the JAXRS implementation, usually already present in the container itself. So there will likely be clashes. We would need to disable the JAXRS impls in the containers (not that hard in WF/EAP, but probably much harder in WAS/WLS). We had similar issues in the past where our rest clients were using resteasy-client dependency. We then decided to rather use the much lighter apache httpclient (but this is probably not an option here as it would need big changes on the Keycloak side). I will take deeper look, but currently I don't really know how to solve this.

 * we are currently using RestEasy 2.x and the Keycloak client seems to need RestEasy 3.x. This may introduce additional incompatibilities as well. I can't really be sure, we would need to test things.

kie-drools-wb/kie-drools-wb-webapp/src/main/webapp/WEB-INF/jboss-deployment-structure.xml
```diff
@@ -51,6 +51,13 @@
       <module name="org.slf4j.ext"/>
       <module name="org.slf4j.jcl-over-slf4j"/>
 
+      <!-- Wildfly / EAP security management provider's client dependencies. -->
+      <module name="org.jboss.as.controller-client"/>
+      <module name="org.jboss.as.domain-management"/>
```
psiroky: According to https://access.redhat.com/articles/1122333, the `domain-management` module is private. This means that there is _no_ backwards compatibility guarantee from the WF/EAP. If we really need to depend on this module, it should be moved few lines below to the private/unsupported section (as the comment suggests).

romartin:
Hi @manstis @psiroky @csadilek ,

Replies to your questions:
- For WAS and WLS there is no user management support, so the perspectives are not visible. See https://github.com/romartin/kie-wb-distributions/blob/user-system-management/kie-wb/kie-wb-webapp/src/main/java/org/kie/workbench/client/KieWorkbenchEntryPoint.java#L173
- Official documentation is not yet done (docbook), but there is some introduction and installation instructions at https://github.com/uberfire/uberfire-extensions/tree/master/uberfire-security/uberfire-security-management
- @psiroky oh yes, it's a private module, but we really need it for managing the properties. So what do you mean by moving the lines below?

psiroky:
@romartin there is a specific section for private modules in the xml file (dom4j is currently there). Would be great to have these private modules at one place and with explanation why we need to depend on them.

romartin:
Hey @manstis! About the menus, @krisv told me to add the perspectives menu entries in the "home" one, and will see where they definitively will end before next community release.... for me np to change it! if we agree :)

romartin:
@psiroky no problem, will move the modules to the right place when agreed on the changes to be done on this PR, thx :)

mdproctor:
Let's leave off the dependencies for now. Instead give instructions that they explode the war, copy the key cloak dependencies and bounce. Additionally we could add an ant.xml or batch file that did this for them, including downloading the dependencies. 

We do not want to end up with permutations of distributions, or bundling stuff to bloat what we have - just in case someone might want it.

romartin:
Okk thx @mdproctor , please @csadilek , @manstis , @psiroky should be nice if we can summarize the missing updates on the PR for the merge, what I understand is:

From my side pending:
- Docbook
- Zanata translations

From the PR comments:
- Move modules on jboss-dep-structure.xml as Petr comments above
- Move menus to "Administration" ??
- For WAS and WLS there is no support (as the initial requirements specify), so the menu entries are not created and the services are null, so not available.
- About the installation instructions, as commented, until not docbook there are some instructions here -> https://github.com/uberfire/uberfire-extensions/blob/master/uberfire-security/uberfire-security-management/README.md. You think are enough for the merge?

Thanks to all!

csadilek:
@romartin proposed changes look good to me.

romartin:
Hi @csadilek , @manstis @psiroky , added a new commit for the proposed changes. Updated too the installation README's while not official docbook. 

About the menu entries, as commented with @mdproctor  and @krisv , as the users management can be used not just for authoring (will see if tasks need some groups, etc), for now let's place in the Home menu and will see the feedback, and in the other hand another refactoring the admin area can affect on next community releases...  

So proposed changes are in the new PR, waiting for your opinions :) Thx!

romartin:
FYI -> Added Keycloak installation instructions to let you know how should it be to install it manually or create some automated task for it -> https://mojo.redhat.com/docs/DOC-1062655

csadilek:
@romartin can you squash the two commits please and resolve the conflicts? 

As mentioned, what I don't like is that there are separate top-menu items for user and group management. IMO, a single user management menu item would be enough allowing me to navigate to group management from there. I don't know why this was decided. Maybe you have some background or I am missing a requirement why these need to be separate?

romartin:
Hi @csadilek, conflicts resolved. commits squashed  and tested all distros again... ready for merge (hopefully) :) Thx

csadilek:
master: 1e668f1775af3546ecfc28d8ed13c46fe6996703

csadilek:
We've merged this so we have a first version available. There are, however, a few open issues:

- KeyCloak setup (see above)
- Menu items (see above)
- Currently when adding or assigning roles/groups in WildFly or EAP a restart is required to activate the changes (creating new users works fine)

/cc @manstis @psiroky

romartin:
@csadilek @manstis @psiroky thx guys! Opened this ticket https://issues.jboss.org/browse/JBPM-4904

PR #174, opened 2016-01-20
JBPM-4906 RestWorkItemHandler fails with CNFE when using ResultSet

agiertli:
Configures the RESTWorkItemHandler with 'classLoader' argument by default, which allows to use ResultClass attribute in the WorkItem without any further changes.


mswiderski:
ok to test

psiroky:
The build is fail is not related to the changes in this PR. Trying to fix the build now.

mswiderski:
ok, thanks @psiroky 

in general +1 to merge as this is config file change only

PR #191, opened 2016-02-09
PLANNER-521- add optaplanner annotations domain editor functionality in kie distributions, but masked out by role plannermgmt

wmedvede:
- Planner domain added to kie-wb, and kie-drools-wb

### DIFF
--- a/kie-drools-wb/kie-drools-wb-webapp/src/main/resources/workbench-policy.properties
+++ b/kie-drools-wb/kie-drools-wb-webapp/src/main/resources/workbench-policy.properties
@@ -45,6 +45,9 @@ feature.wb_release_project=Release Process
 # data modeller features
 feature.wb_data_modeler_edit_sources=Edit Java Sources
 
+# opta planner features
+feature.wb_optaplanner_domain=OptaPlanner Domain
+
 # Groups of features
 # (Features can be excluded by adding the prefix  '!')
 # Groups of project operation
@@ -59,6 +62,7 @@ profile.datasets = wb_datasets
 profile.wb_everything=wb_authoring, wb_search, wb_management, wb_project_operation, wb_plugins, wb_extensions, wb_perspective_editor, wb_apps, wb_datasets, wb_data_modeler_edit_sources
 profile.wb_for_business_analysts=wb_everything, !wb_artifact_repository, !wb_administration, !wb_management, !wb_extensions, !wb_data_modeler_edit_sources
 profile.wb_for_assets_management=wb_configure_repository, wb_promote_assets, wb_release_project
+profile.wb_for_optaplanner=wb_optaplanner_domain
 
 # Granted roles per feature
 # Users with a given role will only be able to access those features specified.
@@ -71,3 +75,4 @@ profile.wb_for_assets_management=wb_configure_repository, wb_promote_assets, wb_
 roles.wb_everything=admin
 roles.wb_for_business_analysts=analyst
 roles.wb_for_assets_management=kiemgmt
+roles.wb_for_optaplanner=plannermgmt
--- a/kie-wb/kie-wb-webapp/src/main/resources/workbench-policy.properties
+++ b/kie-wb/kie-wb-webapp/src/main/resources/workbench-policy.properties
@@ -49,6 +49,9 @@ feature.wb_release_project=Release Process
 # data modeller features
 feature.wb_data_modeler_edit_sources=Edit Java Sources
 
+# opta planner features
+feature.wb_optaplanner_domain=OptaPlanner Domain
+
 # List of project operation 
 feature.wb_project_authoring_save=Project Save Button
 feature.wb_project_authoring_delete=Project Delete Button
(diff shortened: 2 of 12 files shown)


wmedvede:
Hello Pere, would you mind review this PR?

It basically includes the adding of the Optaplanner Dock to the Authoring perspectives for both kie-wb and kie-drools-wb in a server side ApplicationScoped component.

pefernan:
+1

kie-drools-wb/kie-drools-wb-webapp/src/main/resources/workbench-policy.properties
```diff
@@ -45,6 +45,9 @@ feature.wb_release_project=Release Process
 # data modeller features
 feature.wb_data_modeler_edit_sources=Edit Java Sources
 
+# opta planner features
+feature.wb_optaplanner_domain=Opta Planner Domain
```
psiroky: There should not be a space between Opta and Planner.

kie-wb/kie-wb-webapp/src/main/resources/workbench-policy.properties
```diff
@@ -49,6 +49,9 @@ feature.wb_release_project=Release Process
 # data modeller features
 feature.wb_data_modeler_edit_sources=Edit Java Sources
 
+# opta planner features
+feature.wb_optaplanner_domain=Opta Planner Domain
```
psiroky: Same as above. Better to keep the project name consistent.

manstis:
Is kie-wb-distributions exempt from having to write tests? ;-)

wmedvede:
@manstis Mike, this is part of a Perspective configuration and additionally nobody knows what the future of the Docks will be. In my opinion (for this particular case) to try to invent a test for this case just for full filling the bureaucracy  is a waste of time.

psiroky:
This is not directly related to this PR, but will mention it anyway. We really should find a way to share the common stuff between kie-wb-webapp and kie-drools-wb-webapp config/setup code. Currently there is lot of duplication which sometimes results in things being updated only in one place. I know there are things that are only meant for kie-wb and it should be possible to cope with that (e.g. extending the base shared setup/config.)

ge0ffrey:
@psiroky about that DRY violation issue: +1. There's a belief that the new plugin architecture will fix this (and it probably will), but I'd argue it's worth refactoring already now to share as much as possible. This will minimize the work that needs to be done once we migrate to that new plugin architecture.

wmedvede:
@csadilek 

A test was added, so from my side it's ready to merge and I'll not waste more time on this.

Rikkola:
@psiroky @ge0ffrey I think everything needed is already pluggable. Menu's, styles, perspectives and screens. The problem is not just kie-wb and kie-drools-wb. 

Every workbench has the same redundancy for styles, services and so on. A lot could be done. The copy&paste is just legacy from the days when we didn't have time to do these.

Kind of related to the "should we test experimental". Both kie-drools-wb and kie-wb still have this experimental setup that was there just to make them demoable ;)

wmedvede:
@csadilek 
The "1 failing check" above is not related to the stuff pushed in this commit, this are the failing tests:

org.kie.smoke.wb.rest.KieRemoteRestSmokeIntegrationTest.testRestUrlsGroupAssignmentProcess
org.kie.smoke.wb.rest.KieRemoteRestSmokeIntegrationTest.testRestHistoryLogs
org.kie.smoke.wb.rest.KieRemoteRestSmokeIntegrationTest.testRestRemoteApiExtraJaxbClasses
org.kie.smoke.wb.rest.KieRemoteRestSmokeIntegrationTest.testRestRemoteApiRuleTaskProcess
org.kie.smoke.wb.rest.KieRemoteRestSmokeIntegrationTest.testRestRemoteApiHumanTaskProcess

csadilek:
@wmedvede would you mind investigating why these tests fail? It doesn't matter if it's this commit's fault or not. We're one team and these tests are important. We all have to take ownership of these tests. Once you know the root cause, I am sure whoever caused the problem is eager to help.

In contrast to all other tests, we don't want to run the full integration test suite with every commit (it would take too long), so these type of tests will sometimes fail after a merge has happened. Whoever runs into the problem first (you are right now) should investigate (which shouldn't take long).

wmedvede:
@csadilek 

The test seems to run locally and I've tried to find what could be related to other projects but I have not idea of what could be happening. In case it doesn't take long to you to find a clue, it will be very welcome.

psiroky:
I've re-triggered the build on different slave to see if the failures are related to specific slave or not.

wmedvede:
kieremotetests

It seems like the tests worked this time. I've chatted with @mrietveld  this evening and he commented that some weird cases of tests that stopped/started working in the jenkins jobs were also detected for some jobs. But don't know if there are a repeatable use case to check with. Btw, let's see what happens.

csadilek:
6.4.x: 2b65ba7428ab2e1e385d784d601a9d62011dccf7

PR #194, opened 2016-02-12
Stabilize selenium smokes

jhrcek:
Addressing recent flakiness of kie-wb selenium smoke jobs.
1) Problems with navigating to PlugIn Management / Process & Task Dashboard perspectives because the previous perspectives were not fully loaded -> Added explicit wait for those persps. to be loaded.
2) Change in locator after recent i18n commits


psiroky:
+1 to merge. Thanks a lot @jhrcek! (will merge it shortly)

PR #197, opened 2016-02-12
Selenium smoke - create & save asset

jhrcek:
### Work in progress - don't merge yet

Selenium tests that navigates to authoring, creates and saves new enumeration and builds & deploys the current project. Success is only verified by checking that Success notifications appeared.

TODO: it seems that the tests are influenced by some REST API smoke tests, which create some org units and as a result change which org unit is selected by default in the UI. This will have to be dealt with either by isolating the selenium tests or by selecting the correct (Evaluation) org unit / repo / project in project explorer, which would however potentially increase test flakiness.

TODO: there might be another source of possible instability: The new duplicate GAV detection feature displays warning when building & deploying project that's installed in maven repo used by kie-wb. This can be prevented by configuring business central to use its own maven repo (not system default ~/.m2), which would be erased between consecutive builds.


manstis:
Regarding duplicate GAV detection; we could add a System Property to disable the check system-wide (may be useful for some customers too). You can add this to BPMSPL-252 if it is required (I have to make some changes anyway).

jhrcek:
@manstis Good idea. I added it to the jira. Thanks for the tip :)

PR #199, opened 2016-02-16
Do not include empty dirs inside WEB-INF/classes

psiroky:
The previous pattern was leaving empty directories which is both ugly and also complicates testing of the patches a bit. I had to list all the dirs specifically as I could find other way (note that we need to keep ...workbench/drools/client/i18n dir in place as it contains required properties files)

### DIFF
--- a/kie-drools-wb/kie-drools-wb-webapp/pom.xml
+++ b/kie-drools-wb/kie-drools-wb-webapp/pom.xml
@@ -1560,7 +1560,11 @@
         <configuration>
           <packagingExcludes>
             WEB-INF/classes/javax/**/*.*,
-            WEB-INF/classes/**/client/**/*.class,
+            WEB-INF/classes/org/kie/workbench/drools/client/**/*.class,
+            WEB-INF/classes/org/kie/workbench/drools/client/resources/i18n/App*.properties,
+            WEB-INF/classes/org/kie/workbench/drools/client/docks/**,
+            WEB-INF/classes/org/kie/workbench/drools/client/navbar/**,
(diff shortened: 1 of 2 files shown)


manstis:
@psiroky What ```.properties``` files are you more specifically referring? 

I had a quick look at kie-[drools-]wb EAP's WAR and could only see properties files under ```/WEB-INF/classes/org/kie/workbench[/drools]/client/resources/i18n/``` and ```/org.kie.workbench.[drools].KIE[Drools]Webapp/i18n/```.

psiroky:
Those under `WEB-INF/classes/org/kie/workbench[/drools]/client/resources/i18n/`. At least the LoginConstants seem to be needed for the JSP login page?

manstis:
@psiroky Super (yes, you're correct. ```AppConstants``` is *not* needed but since we have GWT source code in ```kie-[drools-]wb-webapp``` I suspect there's little we can do to remove them (other than include a clumsy exclude for ```WEB-INF/classes/**/client/resources/i18n/App*.properties``` (the Java classes are excluded with ```WEB-INF/classes/org/kie/workbench/drools/client/**/*.class```).

manstis:
+1 to merge.

psiroky:
Yeah, these excludes are not ideal indeed. It seems that adding one more for `WEB-INF/classes/**/client/resources/i18n/App*.properties` would make definitely sense. Better to exclude all the stuff we don't need.

Thanks for the additional info @manstis. 

I will amend the PR.

psiroky:
PR updated to include the additional exclusion pattern for AppConstants*.properties.

manstis:
+1 to merge.

PR #204, opened 2016-02-17
Fix for positioning utility menu bar

cristianonicolai:
Moved menu a few pixels down to not overlap with top red bar.

From this: 

selection_057

To:

selection_058


manstis:
+1 to merge.

PR #213, opened 2016-03-02
Add kiemgmt role to WAS web*.xml to make the role available and worki…

zkrejcov:
…ng on WAS


csadilek:
6.4.x: 248b53155418666fdf4dddef330d88ed34c42724

PR #222, opened 2016-03-08
BZ1315760: kie-config-cli: 'Auth failure' trying to work with repositories

manstis:
See https://bugzilla.redhat.com/show_bug.cgi?id=1315760

@csadilek Yes, no tests.. ```kie-cli-config``` is a real can of worms. One day it'll be re-written to use a ```REST``` API (rather than clone ```system.git``` locally...)


csadilek:
master: 58047408a084bfebdae051e15b2234328f03dcdd

PR #233, opened 2016-03-18
JBPM-4937: Data Modeler loading time very long and increasing with more 'instances'

wmedvede:
- Docks optimizations for avoiding unnecessary docks manipulation.

### DIFF
--- a/kie-drools-wb/kie-drools-wb-webapp/src/main/java/org/kie/workbench/drools/client/docks/AuthoringWorkbenchDocks.java
+++ b/kie-drools-wb/kie-drools-wb-webapp/src/main/java/org/kie/workbench/drools/client/docks/AuthoringWorkbenchDocks.java
@@ -59,7 +59,13 @@ public class AuthoringWorkbenchDocks {
 
     private UberfireDock plannerDock = null;
 
+    private String currentPerspectiveIdentifier = null;
+
+    private boolean dataModelerDocksEnabled = true;
+
+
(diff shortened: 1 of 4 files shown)


wmedvede:
@manstis Hello Mike, would you mind review this PR?

This PR should be merged after this one: 
https://github.com/droolsjbpm/kie-wb-common/pull/319

Thanks.

zkrejcov:
Just a side note...the code shown in the diff here looks (on a brief look through) pretty much the same in both kie-wb and kie-drools-wb. Having it in one place, probably kie-wb-common module, seems a bit more appropriate. If there is a bug to fix or any change needed, you would do it in one place only, instead of duplicating the code in two modules (and risking forgetting something).

manstis:
+1 to merge.

@zkrejcov Makes a very good point.. ```kie-drools-wb``` and ```kie-wb``` are really meant to be composition + configuration; and (close to) "no code" (other than bootstrap of CDI producers etc). We probably need to review the "code" in these webapps to avoid duplication. Something @csadilek has also been keen to promote.

csadilek:
@zkrejcov @manstis yes, we need to address code duplication for the WB setup and in parts we already did. My idea was to create a dependency i.e. (wb-setup) that contains CDI managed beans (marked as ```@Alternative```) to set up the default configuration for all WBs (i.e. the ApplicationScopedProducer bean we have everywhere). If needed, these alternatives can be overridden where a customized configuration is needed. This is out of scope for this week's release though. So, I am merging this as is.

csadilek:
6.4.x: 16267cae94791c7adc8edf01c1730091e1180be1

PR #234, opened 2016-03-18
BZ1318739: Product Home Page has localized URLs in Zanata

manstis:
See https://bugzilla.redhat.com/show_bug.cgi?id=1318739

### DIFF
--- a/kie-drools-wb/kie-drools-wb-home-page-product/src/main/java/org/kie/workbench/drools/client/home/HomeProducer.java
+++ b/kie-drools-wb/kie-drools-wb-home-page-product/src/main/java/org/kie/workbench/drools/client/home/HomeProducer.java
@@ -21,6 +21,7 @@ import javax.enterprise.inject.Produces;
 import javax.inject.Inject;
 
 import com.google.gwt.core.client.GWT;
+import org.kie.workbench.common.screens.home.client.widgets.home.HomeImagesHelper;
 import org.kie.workbench.common.screens.home.model.HomeModel;
 import org.kie.workbench.common.screens.home.model.Section;
 import org.kie.workbench.drools.client.resources.i18n.HomePageProductConstants;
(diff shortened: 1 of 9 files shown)


manstis:
@mariofusco @mswiderski Are either of you fixing the compilation failure (I saw you chatting on IRC): https://kie-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/kie-wb-distributions-pullrequests/153/console

mswiderski:
fixed droolsjbpm-integration for both master and 6.4.x

csadilek:
+1

csadilek:
Jenkins please retest this.

manstis:
@csadilek Just for the record, Jenkins won't find any issue with this PR (even if I wrote a class in Urdu). Jenkins doesn't compile the "productized" profile.. so errors in code specific to that profile will never be detected. IDK what other modules else there's specific code for "productized".

@psiroky IDK if it's worth having Jenkins builds for "productized" profile. WDYT?

csadilek:
Test failures unrelated :). However, it does look like we have a problem here. It seems the latest refactoring to use constructor injection in the BasicAuthSecurityFilter doesn't work on Tomcat @mswiderski. Looks like we have to step back to field injection...

csadilek:
master: 138a64b7c446a644e12cee766b0362802da75dfe

psiroky:
@manstis hmm, good question about the "productized" profile. It would definitely be nice, but it would also mean the build time would be even longer (as we would basically need to GWT compile the WB WARs twice). Would need to measure that though.

manstis:
@psiroky Just the webapps? What about ```org.kie.internal.utils.KieMeta.isProductized``` and usage thereof.... I know it's used in one place (in ```kie-wb-distributions``` so that'd be covered anyway) but the flag was added for a good reason and it's likely we have other code behaving differently in "productized".

psiroky:
@manstis so you think we should build + test everything twice for every PR? With and without the `-Dproductized`?

manstis:
@psiroky I'm not proposing anything... I think, given my knowledge of where we have different code paths for "product" vs "community", the effort-to-gain ratio (for building productized in addition to nonProductized) is too low and we're OK as we are (admittedly to be completely thorough and technically correct we should build every PR for every profile... but... well....IDK if we really want to entertain that notion).

PR #245, opened 2016-03-29
[master] BZ1321821 - Add kiemgmt role to WLS web*.xml

zkrejcov:


manstis:
@zkrejcov Thanks for doing the fix (as well as reporting)

+1 to merge.

zkrejcov:
The test fails are in selenium tests, as far as i can tell, they timed-out waiting for the kie workbench to load. Unless these tests run on WLS, this should be unrelated.

psiroky:
The tests are running on Tomcat, so the fails are not related. (although we need to figure out why the tests are failing...)

csadilek:
master: 294fc5c07c3a6b0e4b46d03621e7c41d45918125

PR #255, opened 2016-03-31
kie-config-cli fixes for BZ-1322274 and BZ-1300785

mswiderski:


manstis:
Do you know why we have to fire the SystemRepositoryChangedEvent in kie-cli-config? If it is needed should it not be raised from the services themselves (so, for example, if the Workbench is enhanced to allow groups to be added to Repositories etc we don't need to duplicate the event firing). Also can you explain the use of the System Property?

mswiderski:
@manstis SystemRepositoryChangedEvent is required to keep up to date repository and org unit services as they cache data. So as described in BZ-1300785, when you first list org units then modify their repositories and list units again they are presenting out of date information.
Moreover kie-config-cli works on a disconnected copy of what workbench has though it should behave same, thus a need to fire SystemRepositoryChangedEvent.

When it comes to use of system property, it's just for completeness/correctness to use same user that was used to clone repo to be used as CLIIdentity rather than OS user.

manstis:
@mswiderski Thanks for the explanations.. 

I had a look around and found ```SystemRepositoryChangedEvent``` to be fired already by [ConfigurationServiceImpl](https://github.com/droolsjbpm/guvnor/blob/master/guvnor-structure/guvnor-structure-backend/src/main/java/org/guvnor/structure/backend/config/ConfigurationServiceImpl.java) which is used by (for example) [ProjectService.addGroup()](https://github.com/droolsjbpm/guvnor/blob/master/guvnor-project/guvnor-project-backend/src/main/java/org/guvnor/common/services/project/backend/server/AbstractProjectService.java#L173).

Is the issue therefore that the [WatchService](https://github.com/droolsjbpm/guvnor/blob/master/guvnor-structure/guvnor-structure-backend/src/main/java/org/guvnor/structure/backend/config/watch/ConfigServiceWatchServiceExecutorImpl.java) isn't running properly in ```kie-cli-config```?

We're all well aware of the issues with ```kie-cli-config``` - if it isn't worth any more time investigating why the ```WatchService``` is not executing I can accept this workaround and we can move on...

mswiderski:
it looks like WatcheService is EJB singleton so it's definitely not capable of running in kie config tool. Moreover we explicitly disable daemons for git and ssh https://github.com/droolsjbpm/kie-wb-distributions/blob/master/kie-config-cli/src/main/java/org/kie/config/cli/support/ConfigurationManager.java#L34-L35

manstis:
@mswiderski Great, thanks for the explanations (I wanted to be sure we wasn't committing a fix to the symptom rather than the cause). +1 to merge.

manstis:
master
* https://github.com/droolsjbpm/kie-wb-distributions/commit/f6b2942f974f8d7e6a88968d49b92ad8e532ea3a
* https://github.com/droolsjbpm/kie-wb-distributions/commit/24f0965cc417373ffef007cca3b006aeec3eadd8

PR #257, opened 2016-04-01
BZ1320502: AdministrationPerspective not loading for administrative u…

jesuino:
BZ1320502: AdministrationPerspective not loading for administrative user whose roles is not defined as "admin"

### DIFF
--- a/kie-drools-wb/kie-drools-wb-webapp/src/main/java/org/kie/workbench/drools/client/perspectives/AdministrationPerspective.java
+++ b/kie-drools-wb/kie-drools-wb-webapp/src/main/java/org/kie/workbench/drools/client/perspectives/AdministrationPerspective.java
@@ -34,7 +34,6 @@ import org.uberfire.client.workbench.panels.impl.MultiListWorkbenchPanelPresente
 import org.uberfire.client.workbench.panels.impl.SimpleWorkbenchPanelPresenter;
 import org.uberfire.mvp.Command;
 import org.uberfire.mvp.impl.DefaultPlaceRequest;
-import org.uberfire.security.annotations.Roles;
 import org.uberfire.workbench.model.CompassPosition;
 import org.uberfire.workbench.model.PanelDefinition;
 import org.uberfire.workbench.model.PerspectiveDefinition;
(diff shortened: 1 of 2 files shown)


manstis:
@jesuino -1 to this I am afraid.

Whilst what you've done is part of the solution it simply removes all restrictions to access of the Administration Perspective. What is also needed is to protect access to this Perspective by using ```KieACL``` on the Menus/Home Page sections etc so that access can be restricted by configuration.

I've cc'ed @paulovmr who has been assigned the BZ for his awareness that you are also looking into a fix.. @paulovmr and I spoke briefly on IRC yesterday about what was required. Please liaise with him.

jesuino:
@manstis Yesterday we were in a hurry with this issue, so I thought to speed up by providing a PR just in case this was that simply. I was skeptical this could be the solution as well.
I will talk to @paulovmr  - thanks!

paulovmr:
@manstis Hi, I took a look in this issue and it seems that all menu access to this feature are already protected by KieACL ([kie-wb](https://github.com/droolsjbpm/kie-wb-distributions/blob/master/kie-wb/kie-wb-webapp/src/main/java/org/kie/workbench/client/KieWorkbenchEntryPoint.java#L264) and [kie-drools-eb](https://github.com/droolsjbpm/kie-wb-distributions/blob/master/kie-drools-wb/kie-drools-wb-webapp/src/main/java/org/kie/workbench/drools/client/KieDroolsWorkbenchEntryPoint.java#L263)). Also, the workbench-policy.properties files already cover this feature in their roles.

So I tested @jesuino 's solution and it worked locally for me. Is there anything else we are missing?

manstis:
Great stuff; if use of KIE ACL is already implemented then there is little else to do than remove the Role annotations. Thanks to you both.

manstis:
+1 to merge.

psiroky:
Sorry to intrude, I was trying the new Github feature: https://github.com/blog/2141-squash-your-commits. This is exactly what I was waiting for for the past few months! We can now merge with rebase directly from Github UI!

manstis:
@psiroky Does the squash rebase rather than merge to keep linear history? It'd be lovely to be able to commit PRs via github's UI rather than CLI.

psiroky:
@manstis yes it does rebase. I merged this PR via the UI and there is no merge commit associated with it when looking at https://github.com/droolsjbpm/kie-wb-distributions/commits/master

PR #264, opened 2016-04-06
do not commit - Fix for failing selenium tests

cristianonicolai:


psiroky:
Even 300 seconds does not seem to be enough. Something else is going on. The home page simply does not load.

cristianonicolai:
yeah, got to the same conclusion :)

PR #265, opened 2016-04-06
Fix for failing selenium tests

cristianonicolai:


cristianonicolai:
jenkins retest this

cristianonicolai:
jenkins retest this

cristianonicolai:
jenkins retest this

cristianonicolai:
@psiroky @manstis @csadilek @jhrcek 

Hey guys, looks like this fixes the errors about the integration tests on master. The bad news is that isn't pretty. I'm open to suggestions here. 
So there is actually two issues going on here, first one is related to cdi on Tomcat, which causes the selenium tests to not run on Jenkins. That is fixed by removing the errai-cdi-server jar from the classpath, thanks @csadilek.
Second issue is around rest integration tests. Here what is happening is that the test imports/clone an entire repository and tries to build a project in there. The problem is only visible sometimes because the import rest call returns success but the index still working on the background. That's why it sometimes succeeds and some other it doesn't, depending on if index was completed on time. I'm not sure if there is a way to check that the index operation is done, that's why I moved the thread sleep there which is not pretty but seems to do the trick. Any better way to fix that please let me know.
Related to the discussion on https://github.com/droolsjbpm/kie-wb-distributions/pull/259

kie-wb/kie-wb-distribution-wars/src/main/assembly/component-kie-wb-tomcat-7-common.xml
```diff
@@ -96,6 +96,7 @@
 
             <!-- Errai -->
             <exclude>WEB-INF/lib/errai-jboss-as-support-*.jar</exclude>
+            <exclude>WEB-INF/lib/errai-cdi-server-*.jar</exclude>
```
psiroky: What about kie-drools-wb? Do we need the same exclusion there?

psiroky:
As much as I don't like `Thread.sleep()` it seems to workaround/fix this issue pretty well. Since we really need this working asap, I would be for merging it.

kie-wb/kie-wb-distribution-wars/src/main/assembly/component-kie-wb-tomcat-7-common.xml
psiroky: @csadilek this should be excluded only for Tomcat?

csadilek: @psiroky TL;DR yes.

errai-cdi-server is the 4.x version of errai-weld-integration. We finally renamed the jar as it hasn't been weld-specific in a long time and works with all CDI implementations. However, since the Tomcat build is using an old version of CDI (1.0) and therefore the old errail-weld-integration (3.0.6.Final), we can't have both errai-weld-integration and errai-cdi-server (for CDI 1.1+) on the classpath.

errai-jboss-as-support provids VFS mappings for classpath scanning only needed in WF, JBoss, EAP.

psiroky: Thanks, that is the piece I was missing.

See https://github.com/droolsjbpm/kie-wb-distributions/pull/267 which is about upgrading Weld from 1.x to 2.x. So when that is merged this exclusion should not be needed anymore.

cristianonicolai: @psiroky, adding the exclusion to kie-drools-wb as well.

cristianonicolai:
jenkins retest this

jhrcek:
@cristianonicolai Ahaa, I could have thought of that. In out tests (not that this approach would be recommended/applicable here) we're actually waiting for the message "indexing finished" to appear in the server.log/catalina.out etc. log files.

cristianonicolai:
jenkins retest this

psiroky:
Merged by https://github.com/droolsjbpm/kie-wb-distributions/commit/80d7eead9629c031743601efcde8a966f0537c9d

Thanks @cristianonicolai!

PR #269, opened 2016-04-13
BZ-1326806 - Serialization problems on websphere as a result of BZ-1310739

mrietveld:


mswiderski:
nice one - system properties do magic for all frameworks :)

mrietveld:
Jenkins failure is due to a unrelated smoke test failure. 

@mswiderski or @psiroky, please merge _ASAP_. :)

PR #271, opened 2016-04-15
Prevent occasional fails in selenium smokes on Windows

jhrcek:
Analyzing the occasional fails in [kie-wb-smoke](https://kie-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/kie-wb-smoke-selenium-windows-6.4.x/) on windows it is most likely that navigation to Plugin management perspective fails because Task list perspective was not yet fully loaded (as seen in screenshot artifacts in any of the yellow-ball job runs).

Current implementation was already waiting for perspective loading animation to disappear. Since disappearance of that is the last observable thing that happens in the DOM during the perspective loading, there is nothing better to use for waiting for perspective to load -> I added additional 2s wait for tasklist perspective to load.

### DIFF
--- a/kie-wb-smoke-tests/kie-wb-smoke-tests-main/src/main/java/org/kie/smoke/wb/selenium/model/persps/TasksPerspective.java
+++ b/kie-wb-smoke-tests/kie-wb-smoke-tests-main/src/main/java/org/kie/smoke/wb/selenium/model/persps/TasksPerspective.java
@@ -38,5 +38,7 @@ public class TasksPerspective extends AbstractPerspective {
     public void waitForLoaded() {
         LoadingIndicator indicator = PageFactory.initElements(driver, LoadingIndicator.class);
         indicator.disappear("Loading");
+        // additional wait, because navigation to next persp occasionally fails on windows
+        Waits.pause(2000);
     }
 }


manstis:
@jhrcek Hi.. something you should be aware of: the "loading animation" is only shown (normally) after a client->server side call takes more than 250ms.. If it is quicker it is not shown. So there is *potential* for the "loading animation" to not "disappear" as it was never "shown" in the first place.

manstis:
+1 to merge (when green)

jhrcek:
@manstis that should be ok, because by default selenium is checking the condition every 200ms or so until the predicate is false. If the animation doesn't appear, the waiting ends and the code goes on.

Just to expand my horizons: is this logic for displaying loading after >250ms part of droolsjbpm code? If so, could you point me to where I could find it?

manstis:
@jhrcek Good to hear your tests will be OK. Code for ```BusyPopup``` is [here](https://github.com/uberfire/uberfire-extensions/blob/master/uberfire-widgets/uberfire-widgets-commons/src/main/java/org/uberfire/ext/widgets/common/client/common/BusyPopup.java)

manstis:
+1 to merge.

PR #279, opened 2016-04-25
GUVNOR-2498: Delay when loading top-level menu items for the first time

paulovmr:


manstis:
We'll probably need this (or something similar) in ```kie-drools-wb``` too.

paulovmr:
@manstis You're right, doing now. Thanks!

paulovmr:
@manstis Ok, all done.

manstis:
+1 to merge (when green).

csadilek:
@paulovmr thanks! Is the kieSecurityService call slower on master now or did it just get executed at a later point? Just curious, the fix is fine either way. Although we'd probably need it on optaplanner-wb, livespark, and drools-wb as well.

paulovmr:
@csadilek Actually, I removed the menu creation from the userSystemManager.waitForInitialization call and the delay seemed to have stopped. So I think that the calls made by the ClientUserSystemManager got slower than the calls on the 6.4.x. branch.

Ok, I will make this change in those webapps too.

manstis:
drools-wb, optaplanner-wb, designer-standalone and jbpm-console-showcase (hope I got the names right) are just developer conveniences and hence the change matters less (IMO at least!)

csadilek:
@paulovmr ok thanks. 

An idea I had to get rid of the duplication in these entry points is to add a ```DefaultWorkbenchEntryPoint``` to kie-wb-common. Then all these different web apps would automatically inherit the default configuration and could customize it by subclassing and using ```@Alternative```. Haven't had a chance to try that yet and see how different these entry points really are and whether or not we'd more than one type. Worth investigating though if we find the time....

manstis:
*ALL* the code in the various webapps could be reviewed and consolidated
into kie-wb-common. In an ideal world the webapps would *only* be
configuration and no code (but this is beyond the scope of the JIRA)

csadilek:
@manstis Less important, yes, but still needs to be done, imo. Sure, the refactoring is definitely not part of this JIRA.

csadilek:
Was about to suggest adding a test, but of course that would mean N duplicates of that test as well. I think I will create a JIRA instead for consolidating that logic and writing tests for it then.

paulovmr:
@csadilek You can assign me to this JIRA when you create it (unless you want to do it yourself).

csadilek:
Here's the JIRA: https://issues.jboss.org/browse/GUVNOR-2502

csadilek:
@paulovmr If you want to take a look, you're *more* than welcome too ;). I won't get to it in the next 2 weeks or so, but I am happy to help.

PR #283, opened 2016-05-09
[DROOLS-1136] move droolsjbpm-uber-distribution

psiroky:
See also https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/196


mbiarnes:
+1

PR #296, opened 2016-06-03
changes to make kie config cli work again

mswiderski:
here comes some updates to make kie config cli operational again. There is still one issue with weld exclude in beans.xml which does not seem to work (maybe after change of weld version @psiroky fyi)

At least it does work so it won't block @dgutierr - please give it a try and see if that works.

### DIFF
--- a/kie-config-cli/pom.xml
+++ b/kie-config-cli/pom.xml
@@ -56,6 +56,10 @@
     </dependency>
     <dependency>
       <groupId>org.uberfire</groupId>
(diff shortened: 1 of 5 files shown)


psiroky:
I think we really need at least some smoke tests for the CLI tool. The WELD upgrade can very well be the cause, but without failing test/build it is almost impossible to check that everything works.

manstis:
Closing as we've removed ```kie-cli-config``` from master.

PR #297, opened 2016-06-07
GUVNOR-2504: Move Guvnor REST tests from QE test suite

livthomas:
I have taken KIE Workbench REST tests from QE test suite and added many new test cases. Right now, the only container where they can run is Wildfly 8. I have not added support for other containers since it does not make sense for 7.x. Cargo is used to configure and start this container now but it will be replaced as we change KIE Workbench to be able to run independently.

These tests have discovered many potential problems. All of them are described in [GUVNOR-2542](https://issues.jboss.org/browse/GUVNOR-2542) where we can discuss if they really need to be fixed. The reproducers have Jira annotations so you can easily find them in the code. Just comment out this annotation to see the actual behavior since those tests are ignored unless [GUVNOR-2542](https://issues.jboss.org/browse/GUVNOR-2542) is resolved.

### DIFF
--- a/kie-wb-tests/.gitignore
+++ b/kie-wb-tests/.gitignore
@@ -1,0 +1,11 @@
+target/
+local/
+
+# Eclipse, Netbeans and IntelliJ files
+/.*
+/**/.*
+!.gitignore
+/nbproject
+*.ipr
+*.iws
+*.iml
--- a/kie-wb-tests/kie-wb-tests-rest/src/main/java/org/kie/wb/test/rest/User.java
+++ b/kie-wb-tests/kie-wb-tests-rest/src/main/java/org/kie/wb/test/rest/User.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.wb.test.rest;
+
+public enum User {
+
+    REST_ALL("restAll", "restAll1234;"),
+    REST_PROJECT("restProject", "restProject1234;"),
+    NO_REST("noRest", "noRest1234;");
+
+    private final String userName;
+    private final String password;
+
+    User(String userName, String password) {
+        this.userName = userName;
+        this.password = password;
+    }
+
+    public String getUserName() {
+        return userName;
+    }
+
+    public String getPassword() {
+        return password;
+    }
+
+}
--- a/kie-wb-tests/kie-wb-tests-rest/src/main/java/org/kie/wb/test/rest/client/RestWorkbenchClient.java
+++ b/kie-wb-tests/kie-wb-tests-rest/src/main/java/org/kie/wb/test/rest/client/RestWorkbenchClient.java
@@ -1,0 +1,346 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.wb.test.rest.client;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import javax.ws.rs.core.MediaType;
+
+import org.guvnor.rest.client.AddRepositoryToOrganizationalUnitRequest;
+import org.guvnor.rest.client.CompileProjectRequest;
+import org.guvnor.rest.client.CreateOrCloneRepositoryRequest;
+import org.guvnor.rest.client.CreateOrganizationalUnitRequest;
+import org.guvnor.rest.client.CreateProjectRequest;
+import org.guvnor.rest.client.DeleteProjectRequest;
+import org.guvnor.rest.client.DeployProjectRequest;
+import org.guvnor.rest.client.InstallProjectRequest;
+import org.guvnor.rest.client.JobRequest;
+import org.guvnor.rest.client.JobResult;
+import org.guvnor.rest.client.JobStatus;
+import org.guvnor.rest.client.OrganizationalUnit;
+import org.guvnor.rest.client.ProjectRequest;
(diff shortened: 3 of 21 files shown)


manstis:
@mrietveld Hi Marco.. care to have a look at this PR?

kie-wb-tests/.gitignore
```diff
@@ -0,0 +1,120 @@
...
+.tern-project
+
+# TeXlipse plugin
+.texlipse
+
+# STS (Spring Tool Suite)
+.springBeans
+
+# Code Recommenders
+.recommenders/
```
psiroky: Do we need such exhastive .gitignore? It looks it was automatically generated, so we could clean-it up a bit. (e.g. remove the c++ dev stuff, use just ".idea" to ignore the whole dir instead of listing the all the possible patterns). Very minor thing though.

kie-wb-tests/kie-wb-tests-rest/src/main/java/org/kie/wb/test/rest/User.java
```diff
@@ -0,0 +1,25 @@
+package org.kie.wb.test.rest;
```
psiroky: Missing license header (in other files as well...).

kie-wb-tests/kie-wb-tests-rest/src/main/java/org/kie/wb/test/rest/util/HttpRequest.java
```diff
@@ -0,0 +1,129 @@
...
+        return this;
+    }
+
+    public T get() {
+        return call(() -> request.get(returnType));
+    }
+
+    public T post() {
+        if (body != null) {
+            request.body(contentType, body); // TODO maybe always use JSON for body
```
psiroky: Should we do something about this TODO?

kie-wb-tests/kie-wb-tests-rest/src/test/java/org/kie/wb/test/rest/functional/RepositoryTest.java
```diff
@@ -0,0 +1,269 @@
...
+        repository.setGitURL("git://localhost:9418/" + originalRepo);
+
+        testCreateOrCloneRepository(repository);
+    }
+
+    @Test
+    @Ignore
+    public void testCloneRepositorySecured() {
+        // TODO
+    }
```
psiroky: Should we do something with this TODO?

kie-wb-tests/kie-wb-tests-rest/src/test/java/org/kie/wb/test/rest/functional/RepositoryTest.java
```diff
@@ -0,0 +1,269 @@
...
+        assertThat(actual).isNotNull();
+
+        SoftAssertions assertions = new SoftAssertions();
+        assertions.assertThat(actual.getName()).isEqualTo(expected.getName());
+        assertions.assertThat(actual.getDescription()).isEqualTo(expected.getDescription());
+        assertions.assertThat(actual.getUserName()).isEqualTo(expected.getUserName());
+        assertions.assertThat(actual.getPassword()).isEqualTo(expected.getPassword());
+        assertions.assertThat(actual.getRequestType()).isNull(); //.isEqualTo(expected.getRequestType());
+//        assertions.assertThat(actual.getGitURL()).isEqualTo(expected.getRequestType().equals("new") ?
+//                "git://" + expected.getName() : expected.getGitURL());
```
psiroky: Better to just remove the commented code.

psiroky:
Just a few minor comments, otherwise looks good. Thanks @livthomas!

psiroky:
Looking at the build log I see `ERROR] Failed to execute goal org.codehaus.cargo:cargo-maven2-plugin:1.4.19:start (start-container) on project kie-wb-tests: Artifact [org.kie:kie-wb-distribution-wars:war] is not a dependency of the project. -> [Help 1]`. My guess is that this is because the PR job runs the tests on Tomcat 8, which is not configured for the new modules. Not sure if it makes sense to fix that though, since we will (hopefully) very soon move to WildFly 10, so we should then use that.

livthomas:
@psiroky thanks for the feedback. I have made those changes. When it comes to the failing Jenkins job, you can try it locally and see the tests pass. The ones that are questionable are ignored unless [GUVNOR-2542](https://issues.jboss.org/browse/GUVNOR-2542) is resolved.

Note that this is only the first pull request and I plan to add more tests in the future.

livthomas:
I have updated the test suite to use Wildfly 10 and fixed the issue with the container run twice.

psiroky:
Thanks @livthomas. It would be great if we could follow the test naming conventions, e.g. unit test (surefire) ending with `Test` and integration test ending with `IntegrationTest` or `IT`. Then we could also remove most of the surefire/failsafe config overrides. WDYT?

livthomas:
Done. All test names end with `IntegrationTest` now.

Anyway, I think there is some problem when test suites in modules _kie-wb-smoke-tests_ and _kie-wb-tests_ are run one after another. It seems that Wildfly is not shut down properly in the first module and the tests in the second module fail because they try to connect to this instance but use different credentials.

kie-wb-tests/kie-wb-tests-rest/src/main/java/org/kie/wb/test/rest/client/RestWorkbenchClient.java
```diff
@@ -0,0 +1,346 @@
...
+import org.guvnor.rest.client.RepositoryRequest;
+import org.guvnor.rest.client.RepositoryResponse;
+import org.guvnor.rest.client.TestProjectRequest;
+import org.guvnor.rest.client.UpdateOrganizationalUnit;
+import org.guvnor.rest.client.UpdateOrganizationalUnitRequest;
+import org.kie.wb.test.rest.util.HttpRequestFactory;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class RestWorkbenchClient implements WorkbenchClient {
```
triceo: I've been thinking... this seems like a very cumbersome way to implement a JAX-RS API. Can we not use something like [Resteasy Client API](https://docs.jboss.org/resteasy/docs/3.0-beta-3/userguide/html/RESTEasy_Client_Framework.html) and get rid of most of this code?

livthomas: I was looking at [Resteasy Client API](http://docs.jboss.org/resteasy/docs/3.0.17.Final/userguide/html_single/index.html#RESTEasy_Client_Framework) when I was rewriting our framework. Unfortunately, we depend on RESTEasy 2.x which does not contain this feature. So unless we update to RESTEasy 3.x, it will not be possible to make use of it. But even if we switch to it, we will still need some kind of wrapper around it like we have right now, because otherwise there will be a lot of duplicated code in our tests.

psiroky: We are going to upgrade RestEasy to 3.x withing the next few days. I would be fine with merging this PR as is and then we can look into making it better (possibly via the Client API). WDYT?

livthomas: I agree with merging it now and refactoring it after RESTEasy version update. I am going to add more tests to this module in the next weeks anyway. However, we need to fix the way the tests are run first so the build will not fail.

psiroky: Yes, I am looking into the fails right now. My guess is that both `kie-wb-smoke-tests` and `kie-wb-tests` are currently executed in parallel which won't work (but need to confirm that).

psiroky:
+1

PR #299, opened 2016-06-13
WildFly 10 upgrade + remove EAP 6.4 WARs

psiroky:


pom.xml
```diff
@@ -140,4 +140,24 @@
...
+        <version>${version.org.kie}</version>
+        <type>war</type>
+      </dependency>
+      <dependency>
+        <groupId>org.kie</groupId>
+        <artifactId>kie-drools-wb-distribution-wars</artifactId>
+        <version>${version.org.kie}</version>
+        <classifier>wildfly10</classifier>
+        <type>war</type>
+      </dependency>
```
psiroky: This is already fixed in the kie-bom, see https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/218/commits/eb5afc91fc500af5b0fad4e67bbcd2a1afa79325. I will remove it from here.

manstis:
@psiroky A couple of questions; (1) should we remove the EAP6.4 assembly completely or should it be modified to address EAP7.x (if our 6.4 assemblies are *that* broken then I can see their removal and *later* replacement with 7.x assemblies makes sense), (2) I see you're removed some Arquillian (WF10) related configuration. Is this correct? @mrietveld Could you please review the smoke-test changes too?

psiroky:
@manstis in case we would need separate distro for EAP 7 (different from the WFLY10) one, we should start from the WFLY10 WAR anyway, since there is quite a few changes between EAP 6.4 and EAP 7.

The WildFly 10 config was duplicated in the ARQ tests module, so I just removed the one copy (the diff looks very weird though). If you take a look at https://github.com/psiroky/kie-wb-distributions/blob/b825f6fc115868586b0ca15e4a61eee41250b164/kie-wb-smoke-tests/kie-wb-smoke-tests-arq/pom.xml#L555 the WildFly 10 config is still there.

manstis:
@psiroky Thanks for the explanations. +1 to merge.

psiroky:
Let me clarify the last sentence in the above comment: The POM had config for both wfly8 and wfly10, so I removed the wfly8 one + did some small changes in the wfly10 one.

PR #300, opened 2016-06-14
Remove no longer supported Tomcat, WAS and WLS WARs

psiroky:


psiroky:
@manstis yes, that is my understanding. @csadilek or @etirelli can hopefully confirm that.

etirelli:
@psiroky: for community, we will continue to support Tomcat, but we should upgrade to the latest version. WAS and WLS, in theory will be dropped, but honestly, I think we should keep them around if that doesn't demand much effort right now. Closer to the release we will have stronger evidence and full commitment from the PM to drop WLS/WAS and then we do it.

psiroky:
My understanding was that we will support only WFLY10 (and possible 11 in future) and later wildfly-swarm? I think supporting multiple container is major pain and requires quite a lot of effort which can could be invested elsewhere. My guess is that the WAS/WLS WARs are already outdated since no one really uses/tests them (but really just guessing).

psiroky:
Closing the PR since it won't get merged (not anytime soon anyway).

PR #304, opened 2016-06-28
GUVNOR-2504: Move Guvnor REST tests from QE test suite

livthomas:
The last pull request regarding [GUVNOR-2504](https://issues.jboss.org/browse/GUVNOR-2504) with the following changes:
- refactor RestWorkbenchClient using JAX-RS 2.0 Client API
- fix testDelete() in JobIntegrationTest
- add access control tests with various roles


kie-wb-tests/kie-wb-tests-rest/src/test/java/org/kie/wb/test/rest/RestTestBase.java
```diff
@@ -28,9 +33,11 @@
 
 public abstract class RestTestBase {
 
-    private static final String URL = System.getProperty("kie.wb.url", "http://localhost:8080/kie-wb");
+    protected static final String URL = System.getProperty("kie.wb.url", "http://localhost:8080/kie-wb");
     protected static final String USER_ID = System.getProperty("kie.wb.user.name", User.REST_ALL.getUserName());
-    private static final String PASSWORD = System.getProperty("kie.wb.user.password", User.REST_ALL.getPassword());
+    protected static final String PASSWORD = System.getProperty("kie.wb.user.password", User.REST_ALL.getPassword());
+
+    protected static final String GIT_URL_REMOTE = "https://github.com/droolsjbpm/jbpm-playground.git";
```
psiroky: My concern with this is that the repository may change at any time and the tests could then start failing (and this can be really to figure out). It could also happen that github is down, it which case the test would fail as well. And also there is an issue with productized builds where the build actually does not have access to 3rd party sites, so the tests would likely fail.

Could we create a git repository as a resource on filesystem and use that? I am not saying we need to do it now, just trying to figure if there is a better way (even if implemented in future).

livthomas: I was thinking about some other solution and nothing came to my mind. But a git repository on a local file system sounds good. I will try to implement it using JGit.

manstis:
We do create an offline repository [here](https://github.com/droolsjbpm/kie-wb-distributions/tree/master/kie-wb-example-repositories) which is ultimately still cloned from the remote... but you might be able to use the "last successful" build artifacts in your tests?

psiroky:
Thanks @manstis. That sounds a good idea and we can make sure the example repository can be cloned as well (via the tests).

livthomas:
I have changed the tests to create a git repository on a local file system and use it instead of the remote one on GitHub.

psiroky:
+1. I believe creating new empty repository as part of the tests is fine. We should consider using the example repositories, but that is out of the scope for this PR.

PR #310, opened 2016-07-14
JBPM-5071 - remove of jbpm runtime from workbench

mswiderski:
depends on https://github.com/droolsjbpm/jbpm-console-ng/pull/436


mswiderski:
this test failure is sort of expected as it does it was removed. I'll try to fix the test ...

kie-wb/kie-wb-webapp/pom.xml
```diff
@@ -1989,6 +1926,29 @@
           </execution>
         </executions>
       </plugin>
+      <plugin>
+        <artifactId>maven-resources-plugin</artifactId>
+        <version>2.7</version>
```
psiroky: The plugin version should be removed (it already comes via parent hierarchy).

kie-wb/kie-wb-webapp/src/main/config/standalone.xml
```diff
@@ -0,0 +1,404 @@
...
+        <socket-binding name="ajp" port="${jboss.ajp.port:8009}"/>
+        <socket-binding name="http" port="${jboss.http.port:8080}"/>
+        <socket-binding name="https" port="${jboss.https.port:8443}"/>
+        <socket-binding name="txn-recovery-environment" port="4712"/>
+        <socket-binding name="txn-status-manager" port="4713"/>
+        <outbound-socket-binding name="mail-smtp">
+            <remote-destination host="localhost" port="25"/>
+        </outbound-socket-binding>
+    </socket-binding-group>
+</server>
```
psiroky: Why do we need custom WFLY/EAP configuration? In ideal case we should be able to deploy into clean WFLY installation (unless there is of course no other way).

psiroky:
@mswiderski looks good, just few minor comments.

Also, is it now possible to remove the `kie-drools-wb-*` stuff? It seems to me that after removing the jBPM runtime (the kie-remote stuff), `kie-wb-*` is now the same as `kie-drools-wb`? (we can of course do the removal after this PR gets merged to not make it any bigger, just wanted to ask).

kie-wb/kie-wb-webapp/src/main/config/standalone.xml
mswiderski: it's for hosted mode as it requires additional login module to be registered to smoothly interact with kie servers
https://github.com/droolsjbpm/kie-wb-distributions/pull/310/files/81710c81716cc83028e6d912f4cf3dc2d16ae39d#diff-7ffc6cf965e66eccbb22a2d0cc034081R319

kie-wb/kie-wb-webapp/pom.xml
mswiderski: ok, will remove it

mswiderski:
although the runtime is not there but there are few additional screens in kie wb that does not exist in kie drools wb. But I agree that this should not be a reason to keep them as separate war files. It should be possible to use security mechanism we have now to simply hide them  via some installer config.

kie-wb/kie-wb-webapp/src/main/config/standalone.xml
psiroky: My concern is that it brings certain maintenance burden for us, as we should ideally keep up-to-date with the WildFly we currently use (e.g. when we upgrade to WFLY 10.1 we would need to keep in mind to upgrade the config as well).

Would it be an option to call wildfly-maven-plugin + some CLI commands which would update the default config coming with the wildfly installation? I found https://docs.jboss.org/wildfly/plugins/maven/latest/examples/execute-commands-example.html, but I am not 100 % sure it supports what we need -- but I believe it would be worth trying).

psiroky: Hmm, looking at the plugin docs it says it needs running wildfly, which is not ideal as we would basically need to start the wildfly twice (one for config and second time for the hosted mode).

mswiderski: indeed, I did some investigation and couldn't figure out anything better. We could revisit it again. Maybe we can extend the errai wildfly launcher to do it before the server is started. I think wildfly 10 has offline jboss cli support which might be an option.

kie-wb-tests/kie-wb-tests-gui/src/main/java/org/kie/wb/selenium/model/Persp.java
```diff
@@ -92,7 +92,7 @@
 
     private static final List<Persp<? extends AbstractPerspective>> ALL_PERSPS = Collections.unmodifiableList(Arrays.asList(
             HOME_PAGE, TIMELINE, PEOPLE, PROJECT_AUTHORING, CONTRIBUTORS, ARTIFACT_REPOSITORY, ADMINISTRATION,
-            PROCESS_DEPLOYMENTS, RULE_DEPLOYMENTS, JOBS, PROCESS_DEFINITIONS, PROCESS_INSTANCES, TASKS,
+            RULE_DEPLOYMENTS, JOBS, PROCESS_DEFINITIONS, PROCESS_INSTANCES, TASKS,
```
jhrcek: A minor comment: rename RULE_DEPLOYMENTS constant name to EXECUTION_SERVERS to make it consistent with the new perspective name.

kie-wb/kie-wb-distribution-wars/src/main/assembly/tomcat7/README.txt
```diff
@@ -1,50 +1,11 @@
...
-
-3. Define system properties for btm.root, bitronix config file, JBoss logging provider and others.
+1. Define system properties
 
     create setenv.sh (or setenv.bat) file inside TOMCAT_HOME/bin and add following:
 
-    CATALINA_OPTS="-Xmx512M -XX:MaxPermSize=512m -Dbtm.root=$CATALINA_HOME \
-    -Dbitronix.tm.configuration=$CATALINA_HOME/conf/btm-config.properties \
-    -Djbpm.tsr.jndi.lookup=java:comp/env/TransactionSynchronizationRegistry \
+    CATALINA_OPTS="-Xmx512M -XX:MaxPermSize=512m \
```
jhrcek: Does leaving -XX:MaxPermSize make sense here, if we're java8-only now?

kie-wb-tests/kie-wb-tests-gui/src/main/java/org/kie/wb/selenium/model/Persp.java
mswiderski: updated

kie-wb/kie-wb-distribution-wars/src/main/assembly/tomcat7/README.txt
mswiderski: removed

PR #311, opened 2016-07-19
Fix build: asset-mgmt-project no longer exists

psiroky:


mswiderski:
+1

psiroky:
@mswiderski let's wait for the build to finish as I did not run it locally.

mswiderski:
@psiroky +1 is implicitly bound to build success :)

psiroky:
@mswiderski I wasn't actually 100% sure about that :)

psiroky:
Not sure if the test fail is related to these changes, will try to investigate.

psiroky:
Since it will take me some time to investigate this; and without these changes the build fails, I would vote for merging it asap. In case I found what is wrong with the tests, I will submit new PR.

mswiderski:
+1

I don't see anyway how this test failure could be related to this change...

psiroky:
I see that one of the earlier PR builds was actually successful (before I changed the commit message and which caused re-triggering the build), so the test is likely just unstable. Will merge this.

PR #312, opened 2016-07-25
JBPM-5205: Conflict between plannermgmt and kiemgmt roles and Security management

dgutierr:
Both roles have been removed.

@mswiderski are these changes enough?


mswiderski:
@dgutierr I believe it's ok from kie-wb/kie-drools-wb point of view, though there might be additional logic in guvnor that deal with enabling/disabling managed repository checkbox on new repo/clone repo wizards. 
@wmedvede could you please correct me if I wrong about this :)

wmedvede:
@mswiderski @dgutierr 
yes, there's some logic in the NewRepositoryWizard in order to check if a given user has the kiemgmt role. If the user has the role, then it can create managed repositories, if not then it can just create simple repositories.

And there also existed some logic in the "Repository Structure" screen ( Repository -> Structure ) in order to enable/disable the "configure", "promote" and "release" buttons depending if the user has role kiemgmt or not. But I'm not sure if this buttons survived after the assets management removal. Probalby the Configure and Release button are still there.

mswiderski:
configure repository remained there, other three (build, promote and release) are gone. So I believe we need to remove the check of the roles in there and let any user who has access to create/clone repo to choose if it shall be managed or not. wdyt?

wmedvede:
@mswiderski yes, makes sense, and probably the same for the "NewRepositoryWizard" ? 
The only thing against letting this options by default to whatever user that can clone/create repositories, is that people will start creating managed repos, and branches without knowing what this means. With a specific role at least people is "forced" to read the doc and investigate what is this role and feature for...

wmedvede:
@mswiderski Hello Maciej, we verified this morning (@dgutierr) that the CretateRepository, and ConfigureRepository permissions works as expected. So it believe it's ready to merge.
+1

mswiderski:
excellent, thanks guys!
in that +1 from me as well :)

PR #313, opened 2016-07-26
Remove kie-wb-smoke-tests module

livthomas:
After jBPM runtime has been removed from KIE Workbench ([JBPM-5071](https://issues.jboss.org/browse/JBPM-5071)) the remote runtime tests in this module are not needed anymore. Guvnor REST API tests has also been replaced by much larger test suite in `kie-wb-tests` module ([GUVNOR-2504](https://issues.jboss.org/browse/GUVNOR-2504)) That means `kie-wb-smoke-tests` module can be deleted.

PR #314, opened 2016-07-26
- cleaning up unused/unmaintained kie-wb-smoke-tests-arq module

mrietveld:
@psiroky I started looking at trying to make `kie-wb-smoke-tests-arq` usable, but it's a lot of work -- and I'm not likely to do it. :/ I'm also the only who would use it. 

In light of that, it seems like a better idea to simply remove it. After all, if we do realize that we want it, we can always just revert the commit. 

What do you think?


psiroky:
+1. I believe the same changes are also part of https://github.com/droolsjbpm/kie-wb-distributions/pull/313 (+ it removes the other deprecated test modules as well)

mrietveld:
@psiroky Haha.. oops. Closing in favor of #313

PR #316, opened 2016-08-01
Removing dead code.

Rikkola:
Ran into this today. The code that used to use these is now in kie-wb-common.


manstis:
+1 to merge. Thanks @Rikkola

PR #317, opened 2016-08-03
[DROOLS-1230] remove WildFly 8 WAR

psiroky:
As agreed with @etirelli we don't need to support WildFly 8 anymore (we WildFly 10 WAR instead).


etirelli:
This is correct, for community we typically only need to support the latest versions of community containers.

psiroky:
jenkins retest this

csadilek:
+1

PR #318, opened 2016-08-04
JBPM-5233: Data source management

wmedvede:
@manstis @csadilek 

Hello Mike, Christian,
This is the data sources management PR related to kie-wb-distributions. Basically adds a couple of   compileSourcesArtifact. Would you mind to take a look?

Thanks!!
Walter.

Depends on:
https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/249
https://github.com/droolsjbpm/kie-wb-common/pull/416

### DIFF
--- a/kie-drools-wb/kie-drools-wb-webapp/pom.xml
+++ b/kie-drools-wb/kie-drools-wb-webapp/pom.xml
@@ -1334,6 +1334,8 @@
               <compileSourcesArtifact>org.kie.workbench.screens:kie-wb-common-examples-screen-api</compileSourcesArtifact>
               <compileSourcesArtifact>org.kie.workbench.screens:kie-wb-common-examples-screen-client</compileSourcesArtifact>
               <compileSourcesArtifact>org.kie.workbench.screens:kie-wb-common-workbench-client</compileSourcesArtifact>
+              <compileSourcesArtifact>org.kie.workbench.screens:kie-wb-common-datasource-mgmt-api</compileSourcesArtifact>
+              <compileSourcesArtifact>org.kie.workbench.screens:kie-wb-common-datasource-mgmt-client</compileSourcesArtifact>
 
               <!-- Models for Drools WB -->
               <compileSourcesArtifact>org.drools:drools-workbench-models-commons</compileSourcesArtifact>
--- a/kie-wb/kie-wb-webapp/pom.xml
+++ b/kie-wb/kie-wb-webapp/pom.xml
@@ -1689,6 +1689,8 @@
               <compileSourcesArtifact>org.kie.workbench.screens:kie-wb-common-examples-screen-api</compileSourcesArtifact>
               <compileSourcesArtifact>org.kie.workbench.screens:kie-wb-common-examples-screen-client</compileSourcesArtifact>
               <compileSourcesArtifact>org.kie.workbench.screens:kie-wb-common-workbench-client</compileSourcesArtifact>
+              <compileSourcesArtifact>org.kie.workbench.screens:kie-wb-common-datasource-mgmt-api</compileSourcesArtifact>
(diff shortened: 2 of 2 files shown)


manstis:
@wmedvede Do we need these changes? Runtime Data Source support is only added to Livespark. These webapps have no dependency on Data Sources (unless they're being brought in transiently - in which case they should probably be excluded).

wmedvede:
Hello @manstis 

I've tried to reduce the changes as much as possible in order to keep the data sources management stuff as isolated as possible. But I think we'll still need this compileSourcesArtifact lines .
I'll try to summarize the story just in case a better solution can be found:

The PersistenceDescriptorEditor belongs to the the data-modeler world, and it now using the "DataSourceSelector" in order to facilitate the selection of the data source. (depending on how all this evolves this is something that might change in the future, but the we need for this integration iteration)

So basically in the data modeler module KieWorkbenchDatamodellerClient.gwt.xml file we have:

  <inherits name="org.kie.workbench.common.screens.datasource.KieWorkbenchDataSourceAPI" />
  <inherits name="org.kie.workbench.common.screens.datasource.KieWorkbenchDataSourceClient" />

This is fine, and I originally thought that no other changes will be required. And I expected that the workbenches in general wouldn't need more addings, since they already has the properly dependencies to the data modeler.

But, when I then tried to build e.g. kie-wb-distributions/kie-wb/kie-wb-webapp, the compilation started to blaim about "DataSourceSelector" with the error "did you forget to inherit a required module?"
I really don't see why, since deps to the data moder are set, etc.

Finally since we are not interested in adding more data sources managemet stuff in other workbenches, I found that the minimal change was to add this compileSourcesArtifact lines. 

But I'm happy to revert this PR if we can came up with a better solution.

manstis:
@wmedvede OK. Thanks for the explanation. I see the inheritance in [Data Modeller](https://github.com/wmedvede/kie-wb-common/blob/8e8a27e0b723f154887f615a14fca39635947480/kie-wb-common-screens/kie-wb-common-data-modeller/kie-wb-common-data-modeller-client/src/main/resources/org/kie/workbench/common/screens/datamodeller/KieWorkbenchDatamodellerClient.gwt.xml). 

I can't think of any other solution.. so happy to +1 these PRs.

@csadilek Any ideas before I give them the green light?

csadilek:
@wmedvede @manstis the only way to solve this is to have two GWT modules for data modeller, one with data sources support and one without i.e. ```KieWorkbenchDatamodellerClient.gwt.xml``` and ```KieWorkbenchDatamodellerClientWithDataSources.gwt.xml```. 

Then LiveSpark would use the latter and everything else the former. This way the jar file (kie-wb-common-datasource-mgmt-client.jar), containing types like ```DataSourceSelector```, doesn't have to be on the classpath if data source management is not used.

I think this is worth doing. We have many modules in UF that don't follow this and are causing problems now (when composing modules).

csadilek:
Of course, in addition the ```PersistenceDescriptorEditorPresenter``` can't directly reference a type in kie-wb-common-datasource-mgmt-client.jar (i.e. ```DataSourceSelector```). This can be solved by creating an interface for this type with two implementations, one that is injected by default and one that's injected if kie-wb-common-datasource-mgmt-client.jar is on the classpath. 

This solution we used in a couple of places i.e.: https://github.com/uberfire/uberfire/blob/master/uberfire-extensions/uberfire-preferences/uberfire-preferences-api/src/main/java/org/uberfire/ext/preferences/shared/impl/PreferenceScopeTypesProducer.java#L48

wmedvede:
@manstis @csadilek I've applied the sugestions by creating the two gwt, modules, the producer, etc, and it's wroking well. There's no need to touch all the worbenches, so this PR can be closed.

see https://github.com/droolsjbpm/kie-wb-common/pull/416/files#diff-4691e1aeb357df40e7f6f99ce1923d59

csadilek:
Nice! Thanks @wmedvede

PR #320, opened 2016-08-05
kie-wb-tests-gui: tasks perspective is no longer in kie-drools-wb

psiroky:


manstis:
+1 to merge (when green).. great catch @psiroky!

PR #324, opened 2016-08-18
Make timeouts for testing rest client configurable

jhrcek:
We've has migrated our selenium tests to use this RestWorkbenchClient to perform some basic setup for tests (deleting/cloning testing repo etc.) but it seems that the default 10/60s timeouts for deleting/cloning repo are often not enough for tests running on jenkins.

This PR makes the timeouts configurable while keeping backward api compatibility + adds some logging.

@livthomas please review


jhrcek:
Test failure is unrelated, but I reported the failure in [jira](https://issues.jboss.org/browse/GUVNOR-2640)

livthomas:
+1 I did not think that other test suites would use this client when I was creating it.

PR #325, opened 2016-08-18
[6.5.x] fix uf-security-wildfly exclusions in assembly files

psiroky:
* as part of EAP7 WARs work, the uf-ext modules were renamed
   from *-wildfly-<version> to *-wildlfy8-<version> and the
   exclusion thus did not match those modules which caused issues
   mainly on WebLogic


csadilek:
+1

PR #326, opened 2016-08-18
Add EAP7 community WAR

psiroky:
* the WAR is identical to the WildFly10 one, but uses
   different classifier to make it explicit that we support
   EAP7 as well


csadilek:
+1

PR #327, opened 2016-08-24
GUVNOR-2647: proper error handling on logout.jsp, needed when JSM is enabled on EAP

porcelli:
as discussed with @mswiderski; the less risk operation is just add a try catch, so avoid problems on different containers.


csadilek:
+1

PR #328, opened 2016-08-24
GUVNOR-2647: proper error handling on logout.jsp, needed when JSM is enabled on EAP

porcelli:


kie-drools-wb/kie-drools-wb-distribution-wars/src/main/productized/webapp/logout.jsp
```diff
@@ -5,10 +5,13 @@
...
-  javax.servlet.http.HttpSession httpSession = request.getSession(false);
-  if (httpSession != null) {
-    httpSession.invalidate();
+  try {
+    request.logout();
+    javax.servlet.http.HttpSession httpSession = request.getSession(false);
+    if (httpSession != null) {
+      httpSession.invalidate();
+    }
+  } catch ( Exception e ) {
```
rsynek: I wonder, is it really necessary to just swallow the exception? I understand in some situations there is no better solution. If this is the case, could you please add there a comment explaining the reason?

porcelli: I try to avoid use comments on code as much as possible. In my opinion, the swallow here is `clear` as a fail would create no side effect.
Do you think that a WARN log would be enough?

rsynek: Yes, self-documented code is the best approach. In that case - why not to catch java.lang.SecurityException ? WARN log could be a second part of the solution. I think we should avoid swallowing general exception without knowing what actually happened.

psiroky: Just my $0.02:
I think empty catch block is always bad and I would say it definitely qualifies for a clear comment as to why it's empty.

porcelli:
@psiroky @rsynek tried to update, please let me know what you think now.

porcelli:
please jenkins test it again.

psiroky:
Looks good, thanks @porcelli!

csadilek:
+1

csadilek:
Build error unrelated.

PR #331, opened 2016-08-26
Exclude incorrect version of javassist (not compatible with Java8)

csadilek:
@mbiarnes @psiroky


psiroky:
+1 as this is temporary fix (it also does not make sense to wait for build to finish since it will fail on -droolsjbpm-integration currently :)

PR #333, opened 2016-08-29
 RHBPMS-1521 :Navigation Entry For KIE Server Controller Is Wrongly L…

nmirasch:
…abelled "Rule Deployment"

(2) Rename Rule deployments in menu bar and home page to Execution servers
like on master


krisv:
Does this include the productized versions of the home page?

nmirasch:
It includes the menu option renaming in kie-wb and kie-drools-wb. I don't see any reference to "Rule Deployment" in productized home page, either the image that's describes the deploy section.

csadilek:
+1

PR #350, opened 2016-09-09
[6.5.x] RHBPMS-4145: Language selection is not remembered on relogin in Chrome

manstis:
(cherry picked from commit d451fe27224a1d646af571486ebab21453fbce21)

This also incorporates kie-wb-common commit b91d8db65bfd8fc9667ee001c838444bdf8bd352.

This is the corollary of both https://github.com/droolsjbpm/kie-wb-common/pull/457 and https://github.com/droolsjbpm/kie-wb-distributions/pull/349 for 6.5.x.

### DIFF
--- a/kie-drools-wb/kie-drools-wb-distribution-wars/src/main/productized/webapp/login.jsp
+++ b/kie-drools-wb/kie-drools-wb-distribution-wars/src/main/productized/webapp/login.jsp
@@ -41,7 +41,7 @@
                         <strong><i18n:message key="loginFailed">Login failed: Not Authorized</i18n:message></strong>
                     </div>
                 </c:if>
-                <form class="form-horizontal" role="form" action="j_security_check" method="POST">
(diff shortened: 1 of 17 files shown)


manstis:
@csadilek Mind taking a look at the 6.5.x equivalent too?

csadilek:
@manstis shouldn't there be a PR for kie-wb-common (6.5.x) as well? Did I miss that?

manstis:
@csadilek Nope. On 6.5.x the "logout" command is (duplicated) in kie-wb-distributions (in kie-drools-wb and kie-wb entry points). So both "master" PRs are combined into a single PR (on kie-wb-distributions) for 6.5.x. The difference in code structure is due to the removal of such duplicated code across all webapps. I kept it duplicate on 6.5.x for simplicity.

manstis:
I did comment this PR saying it combined the kie-wb-common PR.... but it'd be easy to miss :-)

csadilek:
@manstis ok thanks. and, yes, I missed that ;).

PR #351, opened 2016-09-09
RHBPMS-631: Login page is not localized - problems with dialect support

manstis:
See https://issues.jboss.org/browse/RHBPMS-631


csadilek:
+1

PR #353, opened 2016-09-09
RHBPMS-4129: Unlocalized 'Please wait. Loading application...' spinner

manstis:
See https://issues.jboss.org/browse/RHBPMS-4129

This PR effectively replaces ```kie-wb.html``` and ```kie-drools-wb.html``` with a ```jsp``` equivalent that can handle the i18n of the "Loading" spinner. When merged I'll email BSIG advising that the "start page" (in IDEA) needs to be changed to the ```jsp``` equivalent. The maven configuration has been updated in the PR.

### DIFF
--- a/kie-drools-wb/kie-drools-wb-distribution-wars/src/main/assembly/tomcat7/WEB-INF/web.xml
+++ b/kie-drools-wb/kie-drools-wb-distribution-wars/src/main/assembly/tomcat7/WEB-INF/web.xml
@@ -32,7 +32,7 @@
 
   <filter-mapping>
     <filter-name>Cache Filter</filter-name>
-    <url-pattern>/kie-drools-wb.html</url-pattern>
+    <url-pattern>/kie-drools-wb.jsp</url-pattern>
   </filter-mapping>
 
   <filter-mapping>
@@ -47,7 +47,7 @@
 
   <filter-mapping>
     <filter-name>Host Page Patch</filter-name>
-    <url-pattern>/kie-drools-wb.html</url-pattern>
+    <url-pattern>/kie-drools-wb.jsp</url-pattern>
   </filter-mapping>
 
   <filter>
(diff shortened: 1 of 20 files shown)


csadilek:
@manstis this reminds me to ask. Do you remember why we're having our own copy of the i18n-taglib? https://github.com/droolsjbpm/kie-uberfire-extensions/tree/master/i18n-taglib

We really need to get rid of that for master. /cc @psiroky

csadilek:
+1, although I really don't like that we have to move to .jsp for our host pages. I don't have a better solution right now though.

manstis:
@csadilek we can probably use a filter on a HTML page instead; but we'd then need to write our own localised literal replacer based on ResourceBundles and invented the wheel again.. happy to consider alternatives though.

manstis:
@csadilek https://www.smashingmagazine.com/2014/06/css-driven-internationalization-in-javascript/ could be worth a try.. the "loading" page only has 2 Strings which we could probably "live with" being held in CSS in the host page itself... if using JSPs smells soooo bad.

manstis:
@csadilek Hmm.. The above link, although showing the potential to use CSS selectors to i18n an HTML page as the ```:lang``` pseudo-selector reads the *browser* locale and our use (requirement) of the query string ignored.. So.. atm JSP seems the only(?) solution.

csadilek:
@manstis yeah don't think we have much of a choice right now. certainly not for 6.5.x. any idea why we forked the i18n-taglib or if there's still a reason that keeps us from using an official one?

manstis:
@csadilek ah yes, I only responded to one of your comments. Unfortunately I don't know much about the origin of the taglib (fork?).. I seem to recall it appearing around the time RH acquired the Polymeter(?) team in Barcelona. Perhaps @porcelli (ok, not in the same team but likes to play with things), or @dgutierr or sr_pere (sorry forgot nick) recall.

porcelli:
can't remember exactly the details... I just recall that I had to use it as is (old version, not available anywhere used by some polymita forms?).
feel free to use a different stuff; v7 is a new world ;)

PR #357, opened 2016-09-15
RHBPMS-4220: Added RoleRegistry and registered principals that come from Keycloak remote logins as roles.

romartin:
Depends on:
- https://github.com/uberfire/uberfire/pull/509
- https://github.com/uberfire/uberfire-extensions/pull/301.


csadilek:
+1

PR #365, opened 2016-09-29
[6.5.x] RHBRMS-2580 - REST API access control does not work on WebLogic

livthomas:
Fix for [RHBRMS-2580](https://issues.jboss.org/browse/RHBRMS-2580) proposed by @mswiderski. It has been tested with Business Central on WebLogic and really resolves the issue.


csadilek:
+1

PR #369, opened 2016-10-04
GUVNOR-2709 : Move client side Verification codes to a separate module

Rikkola:


manstis:
Jenkins please retest this.

manstis:
@Rikkola You may need to sync your ```droolsjbpm-build-bootstrap``` fork with ```master```

kie-drools-wb/kie-drools-wb-webapp/src/main/resources/org/kie/workbench/drools/KIEDroolsWebapp.gwt.xml
```diff
@@ -70,6 +70,9 @@
   <inherits name="org.drools.workbench.models.commons.DroolsWorkbenchModelsCommon"/>
   <inherits name="org.drools.workbench.models.datamodel.DroolsWorkbenchDataModelAPI"/>
 
+  <!-- Drools Workbench Services -->
+  <inherits name="org.drools.workbench.services.verifier.api.VerifierApi"/>
```
manstis: Do we need to inherit this here? If you inherit it in [DroolsWorkbenchGuidedDecisionTableEditorClient](https://github.com/droolsjbpm/drools-wb/blob/master/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/resources/org/drools/workbench/screens/guided/dtable/DroolsWorkbenchGuidedDecisionTableEditorClient.gwt.xml) we should not need it here.

kie-wb/kie-wb-webapp/src/main/resources/org/kie/workbench/KIEWebapp.gwt.xml
```diff
@@ -76,6 +76,9 @@
   <inherits name="org.kie.workbench.common.services.KieWorkbenchCommonServicesAPI"/>
   <inherits name='org.kie.workbench.common.services.refactoring.KieWorkbenchCommonRefactoringAPI'/>
 
+  <!-- Drools Workbench Services -->
+  <inherits name="org.drools.workbench.services.verifier.api.VerifierApi"/>
```
manstis: Do we need to inherit this here? If you inherit it in [DroolsWorkbenchGuidedDecisionTableEditorClient](https://github.com/droolsjbpm/drools-wb/blob/master/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/main/resources/org/drools/workbench/screens/guided/dtable/DroolsWorkbenchGuidedDecisionTableEditorClient.gwt.xml) we should not need it here.

Rikkola:
jenkins please retest

manstis:
@Rikkola just a couple of unanswered questions, then I can merge..

Rikkola:
Going to double check if those are needed. You should be right and they can be removed.

kie-drools-wb/kie-drools-wb-webapp/pom.xml
```diff
@@ -523,6 +523,13 @@
       <artifactId>kie-wb-common-workbench-backend</artifactId>
     </dependency>
 
+    <!-- Drools Workbench Services -->
+    <dependency>
+      <groupId>org.drools</groupId>
+      <artifactId>drools-wb-verifier-api</artifactId>
+      <scope>provided</scope>
```
psiroky: Should this really be `provided`? It seems that the other `-api` modules re not being marked as such. (I may be overlooking something, but it just caught my eye...)

Rikkola:
Removed the provided scope and the imports from gwt.xml files

kie-drools-wb/kie-drools-wb-webapp/pom.xml
psiroky: Sure, no worries, just wanted to double check it is intended (if it would be to stay, it could use a comment explaining the situation...)

PR #370, opened 2016-10-05
Add simple import, build & deploy project test to seleinum test suite

jhrcek:
The "meat" of this PR is `ProjectAuthoringIntegrationTest` which imports Evaluation project from jbpm-playground, does build & deploy (just to m2 repo, no kie-server) and checks in artifact repo perspective that project has been installed. 

Also udpated IEDriverServer version to match selenium version we're using and removed ProcessDeploymentsPerspective which is no longer present in the workbench.

Executed the selenium tests in this module with IE 10x on CI windows machine with no failures.


csadilek:
+1

Thank you!

PR #375, opened 2016-10-12
GUVNOR-2722: i18n: Host JSP's do not use correct Locale when the User has configured multiple

manstis:
See https://issues.jboss.org/browse/GUVNOR-2722


manstis:
@krisv @tsurdilo Want to review?

krisv:
+1
@manstis  does this also have an impact on zanata (or should existing translations for this file still be picked up correctly after rename)?

manstis:
@krisv The PR for ```droolsjbpm-build-bootstrap``` ensures the ```LoginConstants_en.properties``` file is pushed to Zanata as a "source" file and hence the source of translations to the other locales.

PR #377, opened 2016-10-13
[JBPM-5366] kie-wb-tests-rest: no need to depend on kie-remote-jaxb

psiroky:


mswiderski:
+1

though it seems that build is hanging....

psiroky:
I forgot one more fix in optaplanner-wb, which caused the build to fail. Now it should really be all, so let's see if Jenkins agrees.

mswiderski:
@psiroky feel free to merge both remaining when they are finished

PR #379, opened 2016-10-20
JBPM-5378: SSO between workbench and dashbuilder does not work

dgutierr:
Fix + tests.

https://issues.jboss.org/browse/JBPM-5378

I've verified on EAP 7.0.2.  The fix can be considered a workaround until we figure out along the EAP guys what's the root cause.

@manstis Can you please review it?

### DIFF
--- a/kie-wb/kie-wb-webapp/src/main/java/org/kie/workbench/backend/LogoutHelper.java
+++ b/kie-wb/kie-wb-webapp/src/main/java/org/kie/workbench/backend/LogoutHelper.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


manstis:
@dgutierr Hi.. you're missing the fix on ```kie-drools-wb``` ("Business Central"). Can you confirm this is deliberate as Dash Builder is not part of B**R**MS?

dgutierr:
@manstis Exactly! I left out BRMS on purpose since it has nothing to do with dashbuilder.

kie-wb/kie-wb-webapp/src/main/java/org/kie/workbench/backend/LogoutHelper.java
```diff
@@ -0,0 +1,48 @@
...
+    public static final void logout(HttpServletRequest request, HttpServletResponse response) throws ServletException {
+        try {
+            request.logout();
+            HttpSession session = request.getSession(false);
+            if (session != null) {
+                session.invalidate();
+            }
+            // JBPM-5378: In the latest EAP 7.x versions SSO does not works if the cookie is not removed
+            Cookie cookie = new Cookie("JSESSIONID", null);
+            cookie.setPath("/dashbuilder");
```
manstis: My only comment is; what is the guarantee that the web-context for Dash Builder is *always* ```/dashbuilder```? Would it be preferential to allow this to be overridden with a System Property? Otherwise I'm happy.

PR #386, opened 2016-10-28
JBPM-5387: Stunner integration into KIE and Drools Workbenches.

romartin:
Stunner's editor, screens and services Integration into both KIE and Drools workbenches.
- By default the BPMN files are using "bpmn2" as extension, which it's still being used by the jbpm-designer editor.
- The file extensions "bpmn" are now edited by the Stunnerr editor's and other dock screens.
- The title for jbpm-designer resources remains the same -> "Business Process"
- The title for Stunner resources is -> "Business Process (Preview)"
- Both using same resource icon.
- NOTE: Usually there are no ".bpmn" files in our demo repositories, so Stunner should only come into play only when creating a new "Business Process (Preview)" item by scratch.  

See https://issues.jboss.org/browse/JBPM-5387
Depends on https://github.com/droolsjbpm/kie-wb-common/pull/528

Thanks!

### DIFF
--- a/kie-drools-wb/kie-drools-wb-webapp/pom.xml
+++ b/kie-drools-wb/kie-drools-wb-webapp/pom.xml
@@ -1206,6 +1206,204 @@
       <artifactId>uberfire-security-management-wildfly</artifactId>
     </dependency>
 
+    <!-- Form modeler. -->
+    <dependency>
+      <groupId>org.kie.workbench.forms</groupId>
+      <artifactId>kie-wb-common-forms-crud-component</artifactId>
+      <scope>provided</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>org.kie.workbench.forms</groupId>
+      <artifactId>kie-wb-common-forms-api</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>org.kie.workbench.forms</groupId>
+      <artifactId>kie-wb-common-forms-processing-engine</artifactId>
+      <scope>provided</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>org.kie.workbench.forms</groupId>
+      <artifactId>kie-wb-common-dynamic-forms-api</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>org.kie.workbench.forms</groupId>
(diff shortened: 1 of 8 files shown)


romartin:
@pefernan this includes dynamic forms stuff as well.

romartin:
Hi @psiroky , plz can u help? :) you know if this is due to my commit? Not getting this locally, and not sure why it happens yet, don't see any previous error.. thanks in advance! 

Caused by: java.io.IOException: Cannot run program "/opt/tools/jdk1.8.0_102/jre/bin/java": error=7, Argument list too long
01:49:49 	at java.lang.ProcessBuilder.start(ProcessBuilder.java:1048)

pefernan:
@romartin Hi, also added that deps on another PR... let's see who arrives first ;)

psiroky:
@romartin the root cause of the failure is the number of dependency jars we are putting on classpath. Your PR adds quite a few new ones and I think we just hit the threshold even on unix systems. Not sure what to do from the Jenkins side though,  I could try to decrease the workspace path length I guess. However, we need to find a better solution for this.

romartin:
@psiroky thanks for  the update! wow never seen that before... so it's some file descriptors constraint or about concrete "jar" files? An OS restriction or due to the container? ( sorry, again, never seen that before, just asking due to my ignorace :) )

psiroky:
This is I _think_ OS specific. Not sure if there is a way to increase the limit.

csadilek:
@romartin it might make sense to create a combined jar e.g stunner-all.jar etc. to limit the number of jars added to the classpath.

romartin:
@csadilek yes that should be a good idea, I can give it a try! Anyway dont think I'll able to do this today, will have to re-test all stuff and on the other hand this PR was quite urgent to be merged as for wednesday's build, in order to include Stunner on the demo, Lemme know @csadilek @psiroky if you think this can be merged as it is or not, so we can expect it for the demo. thanks!

romartin:
Hi @csadilek , as commented, I have opened [this ticket](https://issues.jboss.org/browse/JBPM-5390) and will start with it as soon as possible. I think it's a must doing this as soon as possible.

csadilek:
+1

PR #388, opened 2016-10-31
RHBPMS-4430: Remove Russian and Chinese (Traditional) from language dropdown

paulovmr:


paulovmr:
Part of an ensemble:

* https://github.com/droolsjbpm/kie-wb-common/pull/533
* https://github.com/droolsjbpm/kie-wb-distributions/pull/388

cristianonicolai:
@paulovmr @manstis I'm wondering if we should also remove the reference to these languages in the [zanata.xml](https://github.com/droolsjbpm/kie-wb-distributions/blob/master/src/main/config/zanata.xml) file across all repos.

paulovmr:
Hi @cristianonicolai , I think we can. @mbiarnes @manstis wdyt? is there any unwanted impact in doing this?

manstis:
@paulovmr I agree with @cristianonicolai the entries in the various ```zanata.xml``` files can go too (it's going to hit every workbench related repository; so you could always just co-ordinate it via BSIG - list repos, assign names and ask for it to be done!?!)

PR #393, opened 2016-11-02
latest Zanata translation changes in kie-wb-distributions from 2016-11-02

kiereleaseuser:


manstis:
@mbiarnes Hi.. I see problems with this PR :(

The translated files for "LoginConstants" include an erroneous "_en" for most locales. This is probably an issue with the Zanata configuration - I am investigating - however perhaps short-term for 6.5.x you could rename files "LoginConstants_en_de.properties" etc (for all locales other than "LoginConstants_en.properties" you could rename to "LoginConstants_de.properties" etc?

Thanks.

mbiarnes:
@manstis DONE

PR #400, opened 2016-11-10
Increase memory for full profile GWT build

csadilek:
Jenkins keeps running OOM when running the full profile GWT build. This started yesterday. I can't reproduce locally (using either profile) but suggest to increase memory of the full profile to match the default profile settings (for now).

@psiroky can you take a look, please?

As a side note, I noticed that we're building 14 different permutations in the full profile. I doubt that this is still necessary, but I need to investigate separately. It seems to me that we can bring the build time down significantly.

### DIFF
--- a/kie-drools-wb/kie-drools-wb-webapp/pom.xml
+++ b/kie-drools-wb/kie-drools-wb-webapp/pom.xml
@@ -1958,7 +1958,7 @@
         </property>
       </activation>
       <properties>
-        <gwt.memory.settings>-Xmx3g -Xms1g -Xss1M</gwt.memory.settings>
+        <gwt.memory.settings>-Xmx4g -Xms1g -Xss1M</gwt.memory.settings>
         <gwt.compiler.localWorkers>4</gwt.compiler.localWorkers>
       </properties>
       <build>
--- a/kie-wb/kie-wb-webapp/pom.xml
+++ b/kie-wb/kie-wb-webapp/pom.xml
@@ -2455,7 +2455,7 @@
         </property>
       </activation>
       <properties>
-        <gwt.memory.settings>-Xmx3g -Xms1g -Xss1M</gwt.memory.settings>
+        <gwt.memory.settings>-Xmx4g -Xms1g -Xss1M</gwt.memory.settings>
         <gwt.compiler.localWorkers>4</gwt.compiler.localWorkers>
       </properties>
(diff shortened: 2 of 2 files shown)


manstis:
@csadilek We have 7 supported locales and 2 supported User Agents; which equals 14 permutations :-(

csadilek:
@manstis thanks for clarify, but I think that's good news, because I don't think we need separate permutations for every locale, at least not for Errai i18n. So, if we step away from uiBinder, wouldn't that mean we only have 2 permutations left?

manstis:
@csadilek Not as such UiBinder but GWT's Messages (i18n) mechanism... but yes if we remove our use of Messages then we'd be left with only User Agent permutations  (hmmmm... I wonder if any right-to-left locals would still require permutations.. but IDK if we have any!)

csadilek:
@manstis well, sure, but I would hope that uiBinder -> Errai UI makes up the biggest part of that work and we don't have too much programmatic UI left, but maybe I am being too optimistic.

But, actually, we can get rid of the 12 permutations right away, even without switching to Errai i18n everywhere by simply adding this to our gwt.xml files (i.e. KIEDroolsWebapp.gwt.xml).

```<collapse-property name="locale" values="*"/>```

Just tested this and it works fine!

@psiroky @mbarkley

manstis:
-Dfull ? Really.. I don't see how we'd get the i18n for the locals?!?!?

csadilek:
@manstis yes, -Dfull. It works fine here. The property above makes sure all translations are put into the two permutations instead of compiling a separate permutation for every single translation.

csadilek:
jenkins retest this (failure unrelated)

manstis:
@csadilek woot.. if tested then +a lot from me :-) --- I'm confident @psiroky won't complain!

csadilek:
Note that this PR only contains the mem increase to (hopefully) make jenkins happy. I will have time on Monday to prepare the other PR (for collapsing permutations).

psiroky:
Thanks @csadilek! I just tried to build with the `<collapse-property name="locale" values="*"/>` and it looks great! The only thing I noticed is that the `*.cache.js` files are bigger - 11.3MiB vs 14.8MiB - that is of course expected as they contain all locales. I would say it is not that bad and it probably should be the reason to not use the `collapse-property`.

PR #401, opened 2016-11-11
latest Zanata translation changes in kie-wb-distributions from 2016-11-11

kiereleaseuser:


manstis:
@mbiarnes Hi.. looks like the "file rename" step in the Jenkins i18n jobs hasn't ran (or been added).

PR #403, opened 2016-11-11
Don't build separate permutation for every locale

psiroky:
Thanks to @csadilek for suggesting this. I believe the decreased build time is a great improvement!


csadilek:
@psiroky @manstis Did another round of tests with this commit and all translations worked fine. Thanks, Petr for preparing the PR.

PR #406, opened 2016-11-15
GUVNOR-2743: Login screen is not loaded when Business Central run on non-suported localizaton of OS

manstis:
See https://issues.jboss.org/browse/GUVNOR-2743

This PR introduces two executions that need to be ran after fetching Zanata translations:

```mvn copy-rename:copy@copy-zanata-files``` and
```mvn copy-rename:rename@rename-zanata-files```


csadilek:
+1

PR #408, opened 2016-11-15
Remove escaping of single apostrophe as these i18n files are not used by GWT and hence do not need escaping

manstis:
Manual fix for 6.5.x i18n issues.

Zanata needs FR locale translated. Zanata team advised.

I have confirmed fetching i18n from Zanata does not overwrite these changes (since Zanata has no translations for these FR files and hence are skipped by the Maven plugin).


krisv:
+1

PR #411, opened 2016-11-16
latest Zanata translation changes in kie-wb-distributions from 2016-11-16

kiereleaseuser:


mbiarnes:
@manstis Can you review this one please? It is the last Zanata translation changes from this night.

PR #412, opened 2016-11-16
JBPM-5397: Stunner - fix classpath issue for kie-wb-common-stunner-bp…

jlindop:
…mn-project dependency

### DIFF
--- a/kie-wb/kie-wb-webapp/pom.xml
+++ b/kie-wb/kie-wb-webapp/pom.xml
@@ -1786,7 +1786,6 @@
     <dependency>
       <groupId>org.kie.workbench.stunner</groupId>
       <artifactId>kie-wb-common-stunner-bpmn-project</artifactId>
-      <scope>provided</scope>
     </dependency>
 
     <dependency>


romartin:
Well seen Jeremy, this artifact is all client side, must be provided. +1 
Thanks!

romartin:
@jlindop @manstis wow guys sorry absolutely my fault here, this MUST NOT be provided ,it's all client side.... sorry.. adding a PR for fixing this, it brokes the distributions runtime...

romartin:
Ok @manstis @jlindop , I see the error, on the bpmn-project artifact there is a server side class, wich is fine sa far as I know, this module provides all necessary BPMN stuff that depend on guvor and wb-commons, both client or server.

So the bpmn-project should be inlcuded on the webapp classpath ( removing the provided scope ), but have to add the proper beans definition for cdi to exclude the client package, isn't it? 

Anyway I have provided [this](https://github.com/droolsjbpm/kie-wb-distributions/pull/415), better having the workench running and stunner without this feature, that borking the container startup...  I will provide, if possible,  a PR in a bit that have prepared and will include and test this. Will let you know. Thanks!!

PR #413, opened 2016-11-17
DASHBUILDE-102: Upgrade dashbuilder-validations to use Errai Validation

pefernan:
Dashbuilder was overriding GWT validation instead of using errai-validation. That was causing issues on kie-wb when other components where trying to user Validations.

Fixing validation dependencies to avoid that issue.

Please merge after https://github.com/dashbuilder/dashbuilder/pull/253

### DIFF
--- a/kie-drools-wb/kie-drools-wb-webapp/pom.xml
+++ b/kie-drools-wb/kie-drools-wb-webapp/pom.xml
@@ -1233,6 +1233,39 @@
       <artifactId>uberfire-security-management-wildfly</artifactId>
     </dependency>
 
+    <!-- Validation -->
+    <dependency>
+      <groupId>org.jboss.errai</groupId>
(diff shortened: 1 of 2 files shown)


pefernan:
@dgutierr @romartin can you review it?

romartin:
@dgutierr , @pefernan Cheking....

romartin:
@dgutierr @pefernan well guys cannot find who can merge this now, so this brokes the build. I have reverted the [commit on dashbuilder for it ](https://github.com/dashbuilder/dashbuilder/pull/253)until both can be merged at same time. Sorry for the inconveniences.

pefernan:
@manstis @romartin @dgutierr the point here is that we should synchronize bot commits on dashbuilder and here, so first do the change on dashbuilder and then merge this.

PR #414, opened 2016-11-17
Removing Russian and Chinese (Traditional) from zanata.xml

paulovmr:


manstis:
@paulovmr Thanks for doing these (an otherwise thank-less task!)

Could you please email BSIG listing all Workbench related repositories and ask people to (a) ensure Zanata website has these locales removed too, (b) ensure any rogue i18n files for these locales are removed (as I see, for example, ```ru``` has crept [back](https://github.com/droolsjbpm/kie-wb-distributions/tree/7771f568ce50a4c050e10c04dc1178ae2f51628b/kie-wb/kie-wb-webapp/src/main/resources/org/kie/workbench/client/resources/i18n) in with the last fetch of translations by @mbiarnes).

PR #423, opened 2016-11-29
Fix failing tests

manstis:


psiroky:
Thanks @manstis! @jhrcek could you please review this?

kie-wb-tests/kie-wb-tests-gui/src/main/java/org/kie/wb/selenium/model/Persp.java
```diff
@@ -51,6 +52,9 @@
     public static final Persp<ProjectAuthoringPerspective> PROJECT_AUTHORING
             = new Persp<>("Authoring", "Project Authoring", ProjectAuthoringPerspective.class);
 
+    public static final Persp<ProjectLibraryPerspective> PROJECT_LIBRARY
+            = new Persp<>("Authoring", "Project Authoring", ProjectLibraryPerspective.class);
```
psiroky: Should we call this Library?

manstis: No, because the Strings are used to identify the menu item on the main navigation menu.

psiroky: Ah, that confused me a bit :) So just to understand. When I go to Authoring -> Project Authoring I may end up on authoring or library perspective? Depending if I have some project already present?

manstis: Yes, if the Workbench does not have a Project then selecting "Authoring" takes you to the "Library".

manstis:
Jenkins please retest this (passes locally for me)

psiroky:
I've tried locally and I am getting same (or similar) failures as Jenkins. Basically the app does not even load:
loginintegrationtest_loginandlogout

manstis:
@psiroky Same here.. looks like the "new failing tests" fail for different reasons; for example:

```
	at org.kie.wb.selenium.model.persps.ArtifactRepositoryPerspective$$EnhancerByGraphene$$f1865954.isDisplayed(<generated>)
	at org.kie.wb.selenium.ui.LoadAllPerspectivesIntegrationTest.allPerspectivesCanBeLoaded(LoadAllPerspectivesIntegrationTest.java:37)
Caused by: org.openqa.selenium.NoSuchElementException: Unable to locate element: {"method":"xpath","selector":"//th[contains(text(),'Name')]"}
```

jhrcek:
@psiroky @manstis to make debugging of issues with selenium tests easier I suggest you set up archiving of screenshots from the test. I already set up screenshots on test failure in the past so it should be only matter of archiving setup in job configuration.
E.g. see messages like these in job's console output for what path should be archived.

Screenshot on failure taken: /home/jenkins/workspace/wb-prs/kie-wb-distributions/kie-wb-tests/kie-wb-tests-gui/target/screenshots/ProjectAuthoringIntegrationTest_importAndBuildProject.png

psiroky:
@jhrcek yeah, its configured, but whatever reason, the screenshots are not archived. I've double checked the paths and it seems be correct. I'll need to take another look to figure out whats wrong.

manstis:
Jenkins please retest this.

manstis:
@psiroky I've confirmed this now works locally using ```mvn clean verify -Pkie-wb,wildfly10 -Dwebdriver.firefox.bin=/home/manstis/installs/firefox/ff46.0/firefox/firefox```. Selenium 2.53.0 requires FireFox 46.0.. is it possible, if Jenkins still complains, that the slave has a newer version of FF?

manstis:
@psiroky I know you're exceedingly busy; but is there any chance you could try running this fix locally with FireFox 46 (using the above CLI command)? You will need to delete .m2's org.jbpm "Evaluation" artifact. This PR runs fine for me locally.

psiroky:
@manstis we can upgrade the firefox, but we haven't upgraded Selenium in months, so I don't understand why it started failing just now.

manstis:
@psiroky I ran this locally with the same environment as Jenkins (maven 3.2.5 and ```mvn -B -T2C -e -fae -Pkie-wb,wildfly10 -Dmaven.test.failure.ignore=true -Dgwt.compiler.localWorkers=1 -Dwebdriver.firefox.bin=/home/manstis/installs/firefox/ff46.0/firefox/firefox -Dselenium.homepage.loading.timeout=120 clean install``` - so the only difference is FF. It passed just fine. Running out of ideas.. but I could try the same FF version as Jenkins - do you know what it is!?

psiroky:
@manstis it runs with firefox 45.3.0

manstis:
@psiroky OK, I finally compiled all of droolsjbpm master and ran the kie-wb-7.0.0-SNAPSHOT-wildfly10 WAR. I could login without problem and was taken to the Home Page (no "Activity not found" type popup). Therefore I have to conclude that the WAR generated by Jenkins is somehow flawed. Can we (temporarily) configure the CI Job to be sequential rather than parallel?

psiroky:
@manstis yeah, we can configure that (but I am afraid it will just stay like that forever).

What I don't understand is why is this happening only now? We have been running parallel builds for years now without this kind of issues, so I would expect some recent change triggered this.

manstis:
Hi @psiroky I have no idea why it's just started to fail and unless we look at when it started failing exactly and look at commits across Errai, Uberfire and any other UI-centric repository we may never know.

At this point my main interest is getting ```kie-wb-distributions``` "green" as there are quite a few pending PRs that have not been merged to other repositories as their PRs relating to ```kie-wb-distributions``` related changes fail. 

If we can prove tests do pass with sequential building it at least tells us there's something nasty happening somewhere else. I'd be more than happy to reconfigure the job to parallel after one iteration as sequential. This PR will either continue to fail or pass.

psiroky:
@manstis so my latest theory is that the issue is somehow related to a recent change I did for maven-compiler-plugin. Before we were always forking the compiler into new JVM, which is very costly and time consuming. So I disabled that. Disabling the forking in combination with parallel builds seems to cause the issue somewhere inside the UF annotation processors (even though different modules would use different classloaders). See https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/323 where I enabled the forking again temporarily, but we should make sure to fix this properly in UF (or wherever the bug is).

jhrcek:
@manstis @psiroky I'm wondering how it's even possible that your build is passing locally. My experience with kie-wb eap7 snapshot builds from the master are quite broken (there are "random" ActivityNotFound modals shown in several perspectives when working with it) which started happening with this week builds if I can tell.

manstis:
@psiroky Was this ran with any changes to maven? 

The single failing test is because of the order the tests were ran in.. which should be easier to fix...

psiroky:
Yes, the last run was the maven "fix".

manstis:
Jenkins please retest this.

@psiroky The tests all pass now locally, independent on order they run.. Did you manage to find a fix for the fork/codegen issue? I saw some PRs from your relating to it.

psiroky:
@manstis those PRs don't fix it, its just something I found along the way. I am still looking into properly fixing this.

manstis:
@psiroky @paulovmr ^^ woot.. want to +1 before it goes red? ;)

paulovmr:
If they now pass as jenkins is saying, +1. :)

PR #430, opened 2016-12-15
[6.5.x][RHBRMS-2709] use Hibernate 4 in WFLY10/EAP7 WARs

psiroky:
Upgrading just core Hibernate libraries which seems safe.
Hibernate Validator was not updated as it causes compatibility
issues with javax.validation, Weld and RestEasy.


psiroky:
@manstis, @mswiderski could you please quickly review this?

mswiderski:
looks ok to me, if I got it right, it's to align it with all other assemblies that do use hibernate 4, correct?

psiroky:
@mswiderski basically yes. As per GSS we don't want to support multiple Hibernate versions for 6.x.

psiroky:
@mswiderski is there anything else required, or can we merge this?

mswiderski:
+1 to merge, please go ahead and merge

PR #434, opened 2017-01-03
[RHBPMS-1622] Set property for very short Errai marshaller class names.

mbarkley:


psiroky:
Out of curiosity, why do we need to set this only for the `productized` profile?

mbarkley:
@psiroky We don't want to obfuscate marshaller names in community builds. It makes stack traces harder to understand and is a deterrent to community feedback. Ideally we will get rid of this completely one day once all our customers are using Windows 10+.

psiroky:
@mbarkley my concern is that this will effectively make the community and product WARs different. I know its just minor thing, but we all (in ENG) use only the community bits when developing. We could easily end up with some hard to debug issues which only manifest itself in product as it would be different.

Considering that we won't be doing any new public releases from this branch, would it make sense to set this also for community bits?

mbarkley:
@psiroky I understand your concern, but I think the risk of this particular difference being the cause of a problem is so small that the benefits of enabling the property for the community build are out-weighed by the burden this could cause when diagnosing other problems during development.

csadilek:
+1. I think obfuscating the class names for everyone (especially all developers) to workaround a windows limitation (which is fixed in windows 10) is too much in this case.

csadilek:
Merging this as it's needed for the release and so that we have a fix in, but if still disagreed let's keep discussing here...

PR #442, opened 2017-01-17
[6.5.x] RHBPMS-4571: It is not possible to deploy business central on IBM WebSphere

karreiro:
See https://issues.jboss.org/browse/RHBPMS-4571

### DIFF
--- a/kie-drools-wb/kie-drools-wb-distribution-wars/src/main/assembly/was8/WEB-INF/web.xml
+++ b/kie-drools-wb/kie-drools-wb-distribution-wars/src/main/assembly/was8/WEB-INF/web.xml
@@ -520,9 +520,6 @@
 minutes. it is now 24 hours. -->
   <session-config>
     <session-timeout>1440</session-timeout>
-    <cookie-config>
-      <http-only>true</http-only>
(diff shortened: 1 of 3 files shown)


karreiro:
@porcelli Could you please take a look?

psiroky:
It would be really great if the commit message explained how does this fix the issue. Why is that config not supported on WAS? Can we do something with that, or are there any alternative solutions? These kind of commit messages really do make life easier when looking at the changes few months from now...

baldimir:
Doesn't this reintroduce security issue from https://issues.jboss.org/browse/RHBPMS-4383 ?

triceo:
@baldimir It does. The issue will never be fixed on WAS.

karreiro:
Thanks for the comment @psiroky. When the "<http-only>" is enabled, the WebSphere does not allow changes on the session cookie, which causes an error on the business central start up. This issue can be avoided by removing the cookie from "Security > Global security > Programmatic session cookie configuration" on the WAS. However, I talked to @porcelli and we decided that would be reasonable to remove the property from WARs generated to WebSphere.

PR #451, opened 2017-01-25
Add EAP7 cargo config

psiroky:


psiroky:
The test failures are very likely not related, because those tests are failing without the changes as well. That being said, we definitely need to fix those before merging this.

psiroky:
Thanks for looking at this @manstis!

manstis:
@psiroky No problem.. all tests were green when I merged :-)

psiroky:
@manstis yeah, I did trigger the build again to see how much random the fails are :) and it worked in this case.

PR #454, opened 2017-01-31
Remove references for extra css files from jbpm-wb

cristianonicolai:
Necessary CSS rules moved into GWT CSS resources instead.


cristianonicolai:
Depends on https://github.com/droolsjbpm/jbpm-wb/pull/617

PR #456, opened 2017-02-01
Re-factor kie-wb UI integration tests

bsremac:
Commit contains the following changes:
- Divide user logging process in two methods:
    1. Just navigate to login page
    2. Submit user credentials and login
- Add Library perspective to the list of perspectives being navigated to in test
- Re-factor logic of navigating to Project Editor
- Re-factor ProjectLibraryPerspective
- Re-factor handling of "pop up" panel with a "glass" element behind it
- Fix false positive scenario in LoadAllPerspectivesIntegrationTest
- Introduce @Before and @After methods in ProjectAuthoringIntegrationTest
- Re-factor deployAndCheckArtifact method in ProjectAuthoringIntegrationTest

### DIFF
--- a/kie-wb-tests/kie-wb-tests-gui/src/test/java/org/kie/wb/selenium/ui/LoadAllPerspectivesIntegrationTest.java
+++ b/kie-wb-tests/kie-wb-tests-gui/src/test/java/org/kie/wb/selenium/ui/LoadAllPerspectivesIntegrationTest.java
@@ -35,6 +35,7 @@ public class LoadAllPerspectivesIntegrationTest extends KieSeleniumTest {
 
     @Test
     public void allPerspectivesCanBeLoaded() {
+        login.getLoginPage();
         HomePerspective home = login.loginDefaultUser();
 
         for ( Persp<?> p : Persp.getAllPerspectives() ) {
@@ -45,11 +46,10 @@ public class LoadAllPerspectivesIntegrationTest extends KieSeleniumTest {
                 AbstractPerspective perspective = home.getNavbar().navigateTo( _p );
 
                 //ProjectAuthoring will direct to ProjectLibrary if there are no Projects in the workbench
-                if ( perspective instanceof ProjectAuthoringPerspective ) {
-                    if ( !perspective.isDisplayed() ) {
-                        LOG.info( "ProjectAuthoringPerspective not displayed. Trying fallback ProjectLibraryPerspective.." );
-                        _p = Persp.PROJECT_LIBRARY;
-                        perspective = home.getNavbar().navigateTo( _p );
+                if (perspective instanceof ProjectAuthoringPerspective) {
+                    if (((ProjectAuthoringPerspective) perspective).isAuthoringDisabled()) {
+                        LOG.info("No projects in Workbench. Redirected to ProjectLibraryPerspective.");
+                        continue;
                     }
                 }
                 LOG.info( "Navigated to perspective '" + _p.getName() + ".." );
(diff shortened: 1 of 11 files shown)


manstis:
Jenkins please retest this.

kie-wb-tests/kie-wb-tests-gui/src/test/java/org/kie/wb/selenium/ui/LoadAllPerspectivesIntegrationTest.java
```diff
@@ -45,11 +46,10 @@ public void allPerspectivesCanBeLoaded() {
                 AbstractPerspective perspective = home.getNavbar().navigateTo( _p );
 
                 //ProjectAuthoring will direct to ProjectLibrary if there are no Projects in the workbench
-                if ( perspective instanceof ProjectAuthoringPerspective ) {
-                    if ( !perspective.isDisplayed() ) {
-                        LOG.info( "ProjectAuthoringPerspective not displayed. Trying fallback ProjectLibraryPerspective.." );
-                        _p = Persp.PROJECT_LIBRARY;
-                        perspective = home.getNavbar().navigateTo( _p );
+                if (perspective instanceof ProjectAuthoringPerspective) {
+                    if (((ProjectAuthoringPerspective) perspective).isAuthoringDisabled()) {
```
jhrcek: I don't understand why we need both isDisplayed and isAuthoringDisplayed in the ProjectAuthoringPerspective. Can't we collapse it to one method somehow?

bsremac: The second method name is "isAuthoring**Disabled**" which serves to determine whether Authoring perspective exists to begin with or the request to this perspective just redirects the user to Library perspective (as no projects are created in/imported to Workbench). The method is unique and tied only to this perspective as only Project Authoring shows this redirection behavior.

"isDisplayed" method, on the other hand, serves to validate whether the content of the Authoring perspective is displayed properly when it does exist. This method is part of the interface which is being implemented by all other perspectives, so if you still think that having two separate methods to distinguish these two scenarios, I can think of alternatives of removing "isAuthoringDisabled" method.
One alternative would be to use "isDisplayed" method of ProjectLibraryPerspective instead of "isAuthoringDisabled", but the logic around the code we are discussing here would be more complex and hence harder to read.

jhrcek: Ok, now I get it. I misread Disabled as Displayed. Your description and implementation now makes sense to me. +1 to merge after removing the unnecessary @Page annotation as we discussed.

manstis:
@bsremac @jhrcek Please advise when this PR has been updated (with the ```@Page``` annotation removed).

bsremac:
@Page annotation removed from use. 
Instead, waitForDisappearance method in BusyPopup class has been made static and as such used directly where needed, as agreed with @jhrcek.

manstis:
@bsremac Please see comments [here](https://github.com/uberfire/uberfire/commit/c1b16f6e956195c096f87d16a665ccbdea30e0db) regarding failures.

jhrcek:
+1 to merge. @bsremac Could you please watch the fix for the issue mentioned by Michel above to be merged and then rerun the PR jenkins job for this?

manstis:
Jenkins please retest this.

PR #458, opened 2017-02-10
Tomcat needs specific listener order to avoid UF shutdown issues

psiroky:
cc @ederign it indeed seems the order of the execution somehow changed. But I am not exactly sure when that happened, since the Tomcat WARs were broken for ages even before this issue.

There is still one thing I don't understand. The UF clean-up happens basically twice - once directly via the Weld destroy() method and the second time using the DisposableShutdownService. This is something worth taking a look at I think. Ideally the clean-up should happen just once.


psiroky:
Not sure who to ask directly, but would be great is someone could review this as it is currently blocking testing on Tomcat.

manstis:
@psiroky Seems a plausible fix.. +1 from me.

psiroky:
@manstis thanks! I would call this a  workaround :) I believe the real fix needs to be done in UF (or the WB config) as we should not be attempting to destroy the "systemFS" twice. However, I am out of ideas as where exactly the issue is, so I've least done this, to at least make the WARs working.

PR #462, opened 2017-02-14
RHBPMS-4583: User can perform process instance specific operations. F…

nmirasch:
…ix error in Perspectives ids referenced
Fixed the reference to perspectives ids. Maybe is a good moment to change the ids itself and avoid the white spaces in that ids, following the same philosophy than the others perspectives .  What do you think?
 
@manstis, @cristianonicolai  can you take a look?


nmirasch:
I close the PR until fix the test, thanks @manstis

PR #463, opened 2017-02-14
RHBPMS-4583: User can perform process instance specific operations. F…

nmirasch:
…ix error in Perspectives ids referenced

Fixed the test without the user permission to access the process instance list and process definitions list.
@manstis Can you review it again? :) 
many thanks!!


kie-wb/kie-wb-webapp/src/test/java/org/kie/workbench/security/WorkbenchACLTest.java
```diff
@@ -100,9 +100,7 @@
     static final List<String> USER_GRANTED = Arrays.asList(
             "perspective.read.SocialHomePagePerspective",
             "perspective.read.UserHomePagePerspective",
-            "perspective.read.Process Definition",
```
cristianonicolai: @nmirasch not sure I understood why you had to remove the process definition and process instance perspectives. I would expect that to be granted to the user role as they are defined in the `security-module-user.properties ` file that you just fixed.

PR #466, opened 2017-02-21
[6.5.x] GUVNOR-2790: Error happens when Execution server profile is used on Weblogic

karreiro:
Removing this reference from WebLogic web.xml:
``` 
<resource-env-ref>		
  <description>Object factory for the CDI Bean Manager</description>
  <resource-env-ref-name>BeanManager</resource-env-ref-name>
  <resource-env-ref-type>javax.enterprise.inject.spi.BeanManager</resource-env-ref-type>
</resource-env-ref>
```

> WLS12c is a fully featured JEE application server and bean manager exists there by default, so this mapping is not needed.

See: https://issues.jboss.org/browse/GUVNOR-2790


manstis:
@karreiro Looks OK to me (based on the JIRA), but the commit to 6.5.x is postponed (based on @etirelli comment) so will not merge atm.

karreiro:
Ok @manstis

karreiro:
@manstis Could you merge this? :-)

PR #474, opened 2017-03-07
[DROOLS-936] fix duplicated classes + enable enforcer rule

psiroky:


psiroky:
@manstis or @cristianonicolai could you please take a look?

PR #476, opened 2017-03-22
Clean-up jboss-deployment-structure.xml files

psiroky:
The deleted deps are either not needed or they are
automatically detected by the WildFly.

Some of the dependencies (like javax.faces)  also break deployment
with WildFly Swarm.


psiroky:
Jenkins retest this

psiroky:
@manstis @mswiderski could you please take a look?

mswiderski:
looks good to me

PR #477, opened 2017-03-23
RHBPMS-4583: User can perform process instance specific operations. R…

nmirasch:
…estored permissions for the role user.
Allow the access to process definitions and process instances perspectives for the role user.

### DIFF
--- a/kie-wb-parent/kie-wb-webapp/src/main/resources/security-module-user.properties
+++ b/kie-wb-parent/kie-wb-webapp/src/main/resources/security-module-user.properties
@@ -16,8 +16,8 @@
 
 role.user.permission.perspective.read.SocialHomePagePerspective=true
 role.user.permission.perspective.read.UserHomePagePerspective=true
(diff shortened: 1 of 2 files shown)


cristianonicolai:
jenkins retest this

cristianonicolai:
@nmirasch should the role user have access to process instance and definition perspectives? If we do allow, then thee is need to block the process operations for this role as mentioned in the JIRA.

nmirasch:
@cristianonicolai I think, the jira was related to add the control and when the user role don't have access to those perspectives, block that link to them. Currently, if the configuration don't want allow the user could perform process operations, the permission have to block those perspectives and the link from task list would appear as disabled. (At least that was what Kris mentioned ) Do you think that makes sense? I understand that we have to modify the role user default permissions in the distribution, but the jira only was referred to react to those permission, and that's the reason for revert this change.

nmirasch:
jenkins retest this

PR #478, opened 2017-03-27
Add WildFly Swarm config to generate workbench uberjars

psiroky:
This is still very far from ideal, but it's a start. The following issues are currently present, but don't block the whole usage of the -swarm.jar, just some parts:
 - https://issues.jboss.org/browse/SWARM-1228 - this issues prevents the datasource integration from deploying jdbc driver via http remoting. Already fixed and will be part of 2017.4.0.
 - https://issues.jboss.org/browse/SWARM-1236 - the workbench app is huge and the deplyoment takes more than the currently set timeout of 30s. This lead to failing tests as the tests are executed before the app is fully deployed. PR sent, will be hopefully merged for 2017.4.0.
 - https://issues.jboss.org/browse/SWARM-1009 - the resulting uberjar is currently huge, more than 350MiB. The dependencies are duplicated. Not sure when will that get fixed.
 - the testing is not ideal either, there is no support for Swarm in the Cargo, so the uberjar is started using the swarm-plugin which leads to some duplication in the test resources. 

@porcelli could you please take a look?


kie-drools-wb-parent/kie-drools-wb-webapp/pom.xml
```diff
@@ -1950,6 +1950,35 @@
...
+            <fraction>cdi</fraction>
+            <fraction>ejb</fraction>
+            <fraction>naming</fraction>
+            <fraction>remoting</fraction>
+            <fraction>jaxrs</fraction>
+            <fraction>ee</fraction>
+            <fraction>security</fraction>
+            <fraction>management</fraction>
+            <fraction>msc</fraction>
+            <fraction>datasources</fraction>
```
psiroky: I will try to remove some these fractions. In theory at least`msc` and `datasources` should be needed.

porcelli:
@psiroky it looks good. I think it's worth wait for 2017.4.0 to merge, wdyt?

psiroky:
@porcelli yeah, we can wait.

porcelli:
@psiroky do we need change anything here for 2017.4.0?

porcelli:
jenkins please retest

psiroky:
@porcelli yes, there is anew config available for 2017.4.0 which we need to increase the defualt deploy timeout (which is 30s and that's usually not enough for the whole WB to deploy).

Once https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/426 is merged, I'll update this one.

porcelli:
@psiroky k, merged ;)

psiroky:
At last, the build is green :) @porcelli I believe this is ready to be merged.

psiroky:
I believe we should _not_ backport this to 7.0.x. It' still a bit experimental. WDYT @porcelli?

porcelli:
@psiroky i'd go for it. @etirelli any objection?

etirelli:
While I think this is isolated enough that would not be a problem to backport, I also think it is probably not worth it as we will be targeting 7.1 for it, correct? I would keep it on master only.

PR #483, opened 2017-04-05
Convert ProjectAuthoringPerspective to ProjectLibraryPerspective

tomasdavidorg:
Library perspective has higher priority so I converted ProjectAuthoringPerspective selenium test to use Library perspective only. Also it should prevent from failing builds in future.


jhrcek:
+1 to merge

psiroky:
Thanks @tomasdavidorg!

Rikkola:
Just noticed this. Thank you @tomasdavidorg this keeps the coverage up. Our plan was to just ignore the tests we break and fix after final.

psiroky:
@Rikkola I will pretend I have not read the part about ignoring the tests :D

Rikkola:
Hehe. Wasn't that happy with it either. We ran all the possible scenarios and this was the result with the time we have. Ignoring or leaving Authoring in just for UI testing. Took my hat off and asked for help. Good to see we got it.

PR #492, opened 2017-04-15
[GUVNOR-2997] jms-api required by the kie-server-client

psiroky:
Backport of #490. @mswiderski I believe we need this for 7.0.x as well, correct?


mswiderski:
That is correct

PR #493, opened 2017-04-17
GUVNOR-3005: Remove search textbox from the menu header

paulovmr:


paulovmr:
@Rikkola Hi Toni, if you could take a look please :)

manstis:
@paulovmr Just one question.. how do Users now perform the "text search" of assets?

Rikkola:
@manstis The Library view, where the assets in the project are listed, has a textbox that allows you to search by file name.
But but, just by file name. If this is a full text search then we need to a search option to that textbox. I know there are several options planned, but 7.0 will only get search by name ( and full text search if this PR removes that functionality ).

manstis:
@Rikkola ok. The ```TextBox``` being removed fronted a "full text" search based on Lucene content (which is sort of "full text" but falls short of what Users might consider "full text"). However if there are plans afoot to change searching for 7.0.x then fair enough.. I'd just comment if text search based on Lucene content is being removed then there's much more "dead code" than this ```TextBox```!

Rikkola:
@manstis @paulovmr Just realized that the search I was talking is per project/repository. Now this was probably for all the repositories. We probably should keep the same screen for searching, but I do see the point of searching from all or from "A and B repository". The question is: how do you access that screen.

manstis:
@Rikkola @paulovmr I've not seen the latest screens; so I can't say with certainty what the current status is; however the "old" Workbench had two ways to search. One was the Text Box in the menu bar - this was "full text" (across the whole workbench) the second was "Asset Search" (scoped to repository) where you'd be taken to a screen to enter meta-data fields.

Rikkola:
@paulovmr @manstis Ok. So based on the wireframe, we can do full text search as an option in the projects list. That lists all the projects.
http://huxhzq.axshare.com/#p=projects&g=1

PR #496, opened 2017-04-18
Disable the WB tests (by default) for the productized profile

psiroky:
@manstis or @mbiarnes please take a look. This is far from ideal, but at least should fix the build for now.


psiroky:
@manstis correct, Jenkins can not test this currently. However, considering the issues we lately had, adding a second run with productized profile may actually be a good idea.

manstis:
Jenkins please retest this.

PR #506, opened 2017-04-21
Disable the WB tests (by default) for the productized profile

psiroky:
Forward port of #496 to make the branches in sync.


manstis:
Jenkins please retest this.

PR #507, opened 2017-04-21
[7.0.x] gui-tests: possibly ovveride GAV conflict when building

psiroky:
@jhrcek or @tomasdavidorg could you please take a look?


jhrcek:
+1 I'm fine with the changes. The alternative would be to disable duplicate gav checking by setting set system property for container running kie-wb `-Dorg.guvnor.project.gav.check.disabled=true`
Pick whatever you find easier ;-)

psiroky:
I was trying to leave the workbench in the default state as much as possible. The check was there before as well it just got removed (probably accidentally) by one of earlier refactorings.

It's true that the tests are note as deterministic as possible as it depends on the local state of the machine the tests are running on. If we find another issues connected to this, I would be fine with disabling the check.

PR #508, opened 2017-04-24
gui-tests: possibly override GAV conflict when building projects

psiroky:
Forward port of #507


psiroky:
@jhrcek could you please (formally) approve this? It is the same change as for 7.0.x.

jhrcek:
+1

PR #509, opened 2017-05-02
kie-wb-tests: Enable deployable version config

livthomas:
This will allow us to configure deployable artifact version when running tests with QE automation using Maven Invoker Plugin.

### DIFF
--- a/kie-wb-tests/pom.xml
+++ b/kie-wb-tests/pom.xml
@@ -24,6 +24,7 @@
     <deployable.classifier.tomcat8>tomcat8</deployable.classifier.tomcat8>
     <deployable.classifier.wildfly10>wildfly10</deployable.classifier.wildfly10>
     <deployable.classifier.eap7>eap7</deployable.classifier.eap7>
+    <deployable.classifier.swarm>swarm</deployable.classifier.swarm>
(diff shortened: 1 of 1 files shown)


kie-wb-tests/pom.xml
```diff
@@ -52,6 +53,60 @@
...
+        <version>${deployable.version}</version>
+        <classifier>${deployable.classifier.swarm}</classifier>
+      </dependency>
+      <dependency>
+        <groupId>org.kie</groupId>
+        <artifactId>kie-wb</artifactId>
+        <version>${deployable.version}</version>
+        <classifier>${deployable.classifier.wildfly10}</classifier>
+        <type>war</type>
+      </dependency>
```
psiroky: I am probably missing something here. @tlivora could you please clarify why we need these? The versions of `kie-wb` WARs should already be in the kie-parent.

livthomas: We are invoking this from QE's test suite and we need to be able to modify the versions to use productized artifacts. Because of this configuration option missing and at the same time wrongly produced artifact versions by the productization we are now not able to run this test suite at all with DR1.

PR #510, opened 2017-05-02
GUVNOR-3046: Fix and re-insert the search box at the header

paulovmr:


paulovmr:
Part of an ensemble:

* https://github.com/kiegroup/kie-wb-common/pull/834
* https://github.com/kiegroup/kie-wb-distributions/pull/510

PR #513, opened 2017-05-04
JBPM-5660: Cloning repositories is allowed even if create repositories permission is disabled (7.0.x)

dgutierr:
Fix + test.

@manstis Can you review it please?


dgutierr:
@manstis Changes done. Thanks for the feedback!

manstis:
Thanks for the changes @dgutierr

PR #514, opened 2017-05-04
Unify Plugin Management and Artifact Repository names in drools-wb

tomasdavidorg:
This unifies Plugin Management and Artifact Repository menu items between kie-drools-wb and kie-wb which breaks qe seleniums tests. 

This is followup on https://github.com/kiegroup/kie-wb-distributions/commit/5df932f687074299dcf986935782b0da092bb754 which changed some item names in top level menu.


jhrcek:
I wonder why this wasn't caught by selenium tests. @psiroky aren't we running those tests with both kie-wb and kie-drools-wb?

psiroky:
We only run the kie-wb tests on WildFly10 now. I agree it would be great to run also the kie-drools-wb tests. It would also be great to run them on all supported containers. That is the only way to make sure we don't break the Tomcat WAR all the time.

PR #515, opened 2017-05-04
[6.5.x] RHBPMS-4706: [GSS](6.4.z) Embedded Authoring perspective is broken when using parameter controlled URL

manstis:
**=== DO NOT MERGE ===**

See https://issues.jboss.org/browse/RHBPMS-4706

@cristianonicolai Would you mind taking a look (at the latest comments on the JIRA and the change).

The change is only for ```kie-drools-wb``` at the moment; but you get the idea..


manstis:
Not needed..

PR #516, opened 2017-05-05
Move Swarm packaging to Community profile

manstis:
@mbiarnes This moves Swarm's plugin execution to Community only profile.

@psiroky FYI.. there is an issue with Swarm's plugin and Maven classifiers on the WAR ("-redhat" for product)


manstis:
@psiroky FYI https://issues.jboss.org/browse/SWARM-1332

psiroky:
@manstis good point, thanks for the workaround and for the SWARM jira.

PR #517, opened 2017-05-08
JBPM-5963 - DMN assets fails to build in workbench

mswiderski:
@etirelli @tarilabs here are the changes to allow DMN assets to be built and deployed from within workbench.

### DIFF
--- a/kie-drools-wb-parent/kie-drools-wb-webapp/pom.xml
+++ b/kie-drools-wb-parent/kie-drools-wb-webapp/pom.xml
@@ -205,6 +205,38 @@
       <artifactId>drools-workbench-models-test-scenarios</artifactId>
     </dependency>
 
+    <!-- org.kie.dmn -->
+    <dependency>
+      <groupId>org.kie</groupId>
+      <artifactId>kie-dmn-api</artifactId>
+      <scope>runtime</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.kie</groupId>
+      <artifactId>kie-dmn-core</artifactId>
+      <scope>runtime</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.kie</groupId>
+      <artifactId>kie-dmn-feel</artifactId>
+      <scope>runtime</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.kie</groupId>
+      <artifactId>kie-dmn-validation</artifactId>
+      <scope>runtime</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.kie</groupId>
+      <artifactId>kie-dmn-model</artifactId>
+      <scope>runtime</scope>
+    </dependency>
+    <dependency>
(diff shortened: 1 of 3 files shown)


tarilabs:
Thanks!

manstis:
This, IMO, brings an interesting concept into question. Projects containing DMN in the Workbench will compile now as the necessary DMN libraries will be on the container's classpath. However said KJARs may not execute on Kie Server when deployed unless either Kie Server includes these dependencies or the KJAR itself declares them in it's POM (in the latter case this PR is not needed)... The same applies for XLS Decision Tables and probably XLS Score Cards etc and their required dependencies. 

This questions reminds me of https://issues.jboss.org/browse/GUVNOR-2274 where we propose the User elects to have the Workbench support different "facets" (Rules, Processes, OptaPlanner, DMN etc). The Workbench then managing the necessary dependencies.

What are your thoughts on the possible impact of DMN assets on Kie Server, Workbench "dependency management" and more generally how we should be handling "additional" dependencies for different asset types, (```kie-api``` for KIE annotations used in Data Modeller, ```drools-decisiontables```, these DMN libraries etc). If we keep including things in the container the KJAR will not compile from source at the CLI.

mswiderski:
DMN execution is already supported by KIE Server via it's dedicated DMN kie server extension so all these libs are already there. Actually this issue was discovered because we already have support for runtime but we lack it on build time.

etirelli:
@mswiderski Thank you Maciej!

@manstis actually the kie-server already supports that. Duncan demoed this integration last week at summit, but mentioned the problem of the error that the workbench was raising.

manstis:
@etirelli @mswiderski Thank-you.. however you perhaps missed; or I poorly posed my question: what should the workbench and runtime be doing in these scenarios!?!? It sounds as though Kie Server has at least some extension point for different runtime requirements; however a KJAR authored in the Workbench will fail to compile from the CLI unless the (for example) DMN dependencies are added manually. What are your thoughts on the Workbench "facet" thing I spoke of!? Does Kie Server discover runtime extensions required for a KJAR; or are they "always enabled"? Would Kie Server need extensions of a KJARs pom was correctly configured? Just thinking of the bigger picture for "feature" vs "dependency" resolution for KJARs.

mswiderski:
I think as long as it has kie-maven-plugin there is no dependencies required to build it manually - from CLI.
When it comes to dependencies. Kie Server (and I think workbench should work the same) provides all dependencies on server class path so projects can be easily deployed to that/build on it. If really needed dependencies to:
- org.kie
- org.drools
- org.jbpm
- org.optaplanner
must always be marked as scope provided.
that's how I see this long term ... and actually short term as well :)

tarilabs:
I agree with @mswiderski as from CLI / kie-maven-plugin normally in the KJAR we would have such dependencies actually with `test` scope as they would be needed for compiling and running the unit tests, but not to be bundled in the kjar.

PR #518, opened 2017-05-08
[7.0.x] JBPM-5963 - DMN assets fails to build in workbench

mswiderski:
@etirelli @tarilabs here are the changes to allow DMN assets to be built and deployed from within workbench - for 7.0.x branch


mswiderski:
tested both kie-wb and kie-drools-wb on WildFly 10 though confirmed all required jars are on all distributions (wildfly, eap, tomcat)

tarilabs:
Thanks!

PR #524, opened 2017-05-09
GUVNOR-3060: Only admins should be able to see the SecurityManagementPerspective

paulovmr:


manstis:
Jenkins please retest this (appears a timing issue.. which should ideally be fixed!)

PR #533, opened 2017-05-30
[6.5.x] RHBPMS-4687: Update the REST API documentation to fix the 'RestApiJspTest'

karreiro:
See: https://github.com/kiegroup/droolsjbpm-integration/pull/991


karreiro:
@krisv Could you please take a look? :-)

PR #535, opened 2017-06-06
JBPM-6063: Workbench - Enable Users Management for the productized profiles.

romartin:
Hello,

This commit enables the users management for the productized profiles (it was actually disabled). See [the JIRA ticket](https://issues.jboss.org/browse/JBPM-6063).

It fixes what I'm pretty sure it's a bug as well - the eap7 maven assembly descriptor for the `kie-wb` was not using the property `webapp.war`, so in fact the productized version of the `kie-wb` was using the regular war as source, not the `redhat` one.


porcelli:
just for the record: i don't think this is a real solution for user management (properties files are not supposed to be used, ever! in production).

psiroky:
Looks good from the Maven point of view. I can not comment on whether this should or should not be enabled by default for the productized WARs.

romartin:
Hey @manstis @porcelli @psiroky thanks for the review. yeah we know that this is not a real production ready setup., Anyway @krisv  asked for this, I'm agree as well, so that's better than do not having it enabled - empty screens on the admin section.

romartin:
Hey @manstis @porcelli @psiroky , 
Quick reminder - please do not forget this one, @krisv need this as soon as possible. I think it's approved so just it's missing to merge it.. :)
Thanks in advance!

romartin:
Thanks!!

PR #540, opened 2017-06-14
GUVNOR-3197: Enable findbugs.failOnViolation

jomarko:


manstis:
@psiroky I'm fine with this PR; however wanted to at least make you aware of it (if you're not already). IDK if the analysis is already performed and this change just means the build could fail; or whether the flag also enables analysis - and hence longer build times (that I know you hold close to your heart).

jomarko:
@manstis yes, it is already executed, just the results are ignored. One example, in this PR I added just tests https://kie-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/drools-wb-pullrequests/725/ and as you can see, findbugs was run.

PR #544, opened 2017-06-29
RHBPMS-4794 - History REST endpoints not behaving correctlly on WAS

mswiderski:


mbiarnes:
Jenkins, please retest this.

mswiderski:
@krisv @mbiarnes slightly improved the test to include items that don not start with [ as is the case for relaxed regex for variable value

PR #548, opened 2017-07-11
[GUVNOR-3280] enable Errai bus CSRF protection

psiroky:
@mbarkley could you please confirm this is all we need?


psiroky:
@mbarkley thanks for the details. Just to double check, by host page you mean the `kie-wb.jsp` which gets loaded after login?

PR #550, opened 2017-07-11
[GUVNOR-3178] new WB wars without kie-server controller

psiroky:
@porcelli could you please take a look? I don't particularly likes this as it adds another 500MiB to our deployables. On the other hand, I did not find better alternative.


psiroky:
Closing as agreed with @porcelli. Adding new WARs is a bad idea. Making the embedded controller optional should be a better way.

PR #553, opened 2017-07-18
DO NOT MERGE: Add standalone perspectives Selenium test.

kkufova:
Hello @manstis,
I created this PR in cooperation with @jhrcek and there are several issues.
The main part of this pull request is a new Selenium test that tries to load each perspective in the standalone mode (both with and without `header=AppNavBar`) and verifies whether the perspective was in fact loaded.

In addition to that, with @jhrcek 's approval, I removed some unused functionality from the tests.

There are several problems that need to be addressed before merging this pull request.

1. In the `loadPerspectiveAndCheckLoadingTime` method, there is commented code that measures the time it takes to load a particular perspective. This part often fails, as some perspectives take more than 30 seconds to load. I am not sure how to identify the cause, however, I believe it should be fixed.
2. In the test itself, you can see I am checking 27 of 29 existing perspectives. Are all of them relevant for 7.x or should some of them be removed from the kie-wb? The list of the perspectives we are unsure about follows:

- _Asset Management_: AFAIK this feature was removed in 7.x. Shouldn't this perspective be removed as a part of a cleanup?
- _AuthoringPerspective_ and _AuthoringPerspectiveNoContext_: Are these two perspectives still relevant now that we have the Library?
- _FormDisplayPerspective_: I am not sure what this perspective does exactly.
- _PreferencesCentralPerspective_: This perspective does not seem to work in the standalone mode.
- _SecurityManagementPerspective_: This perspective is not accessible in the top menu, but we probably still want to test it as a standalone perspective.
- _StandaloneEditorPerspective_: I am not sure how to test this; I would probably need some additional parameter with a path to an asset?
- _WiresGridsDemoPerspective_ and _WiresTreesPerspective_: These two perspectives look more like some kind of a demo; are we sure we want to include them in the product?

3. Is there something more you would like to test?


kie-wb-tests/kie-wb-tests-gui/src/test/java/org/kie/wb/selenium/ui/StandalonePerspectivesIntegrationTest.java
```diff
@@ -0,0 +1,121 @@
...
+            put("TaskAdmin", By.className("qe-list-bar-header-Task-Administration-List"));
+            put("Tasks", By.className("qe-list-bar-header-Tasks-List"));
+            put("UserHomePagePerspective", By.className("fa-home"));
+            put("WiresGridsDemoPerspective", By.className("qe-list-bar-header-Grids"));
+            put("WiresTreesPerspective", By.className("qe-list-bar-header-Palette"));
+        }
+    };
+
+    @Test
+    public void loadStandalonePerspectives() {
```
triceo: Any chance we could make this test `@Parameterized`? To test 35 different perspectives from within one test case seems evil.

Alternatively (as I'm sniffing potential problems with the annotation and Arquillian.), can we at least make the looped code its own function and put some logging around it, so that we know when was it that the test actually failed?

Of course, we could also have a different method for each of the perspectives - which would be the "cleanest" solution in this situation. But I sort of see how that would be overkill.

kie-wb-tests/kie-wb-tests-gui/src/test/java/org/kie/wb/selenium/ui/StandalonePerspectivesIntegrationTest.java
```diff
@@ -0,0 +1,121 @@
...
+        login.getLoginPage();
+        login.loginDefaultUser();
+
+        for (Map.Entry<String, By> entry : perspectiveToElement.entrySet()) {
+            String perspectiveId = entry.getKey();
+            By elementInPerspective = entry.getValue();
+
+            loadPerspectiveAndCheckLoadingTime(perspectiveId, true);
+            verifyPerspectiveIsLoaded(perspectiveId, elementInPerspective);
+            verifyPresenceOfHeader(true);
```
triceo: From the point of test design, `headerIncluded` is definitely two different test cases, one for `true` and one for `false`.

kie-wb-tests/kie-wb-tests-gui/src/test/java/org/kie/wb/selenium/ui/StandalonePerspectivesIntegrationTest.java
jhrcek: Unfortunately in this case we can't use `@Parametrized` as it requires different JUnit Runner. Since we're using `@Drone` to setup WebDriver, we're forced to use `@RunWith(Arquillian.class)` in the [parent test class](https://github.com/kiegroup/kie-wb-distributions/blob/master/kie-wb-tests/kie-wb-tests-gui/src/main/java/org/kie/wb/selenium/model/KieSeleniumTest.java#L28).

There's another issue why we went for 1 test with for loop: the need for `@BeforeClass` setup (logging in once before starting to load all perspectives), which is impossible with the current version of Drone, which is not initialized at the point when `@BeforeClass` is running. I heard this has been fixed in Drone 2 but didn't want to do too many changes at once. We'll do Drone/Graphen/Selenium version update (and related Firefox update in CI images) next. Larger refactoring will be needed anyway once the Megamenu proposal is implemented.

Take this initial implementation as a medium for communication with some details to be worked out:
* Which perspectives do we want to test in this way?
* Do we want to remove some persps from the workbench?
* What do we do with perspective loading times sometimes exceeding 30 seconds?
* Do we want to test more stuff than that the perspective was loaded?
* `header=AppNavBar` only shows useful stuff in AuthoringPerspective, AuthoringPerspectiveNoContext and AdministrationPerspective, so we probably don't need to check the effect of this parameter in other perspectives, right?

 After @manstis provides some clarification, we'll refactor this to "1 test per perspective" as you suggested, which although it requires more boilerplate, will give the benefit of more granular coverage and not have the disadvantage of "fail cause the following things not to be tested".

kie-wb-tests/kie-wb-tests-gui/src/test/java/org/kie/wb/selenium/ui/StandalonePerspectivesIntegrationTest.java
```diff
@@ -0,0 +1,121 @@
...
+        }
+    }
+
+    private void loadPerspectiveAndCheckLoadingTime(String perspectiveID, boolean headerIncluded) {
+        //long loadingStart = System.currentTimeMillis();
+
+        String standalonePerspectiveURL = getStandalonePerspectiveURL(perspectiveID, headerIncluded);
+        driver.get(standalonePerspectiveURL);
+
+        BusyPopup.waitForDisappearance();
```
jhrcek: @kkufova Waiting for `BusyPopup` is useles, as it doesn't appear before `ApplicationLoadingPopup` disappears. Let's remove this line (or move it after App loading popup).

kie-wb-tests/kie-wb-tests-gui/src/test/java/org/kie/wb/selenium/ui/StandalonePerspectivesIntegrationTest.java
```diff
@@ -0,0 +1,121 @@
...
+        //assertThat(perspectiveLoadingTime)
+        //        .as("It should take less than "+PERSPECTIVE_LOADING_TIMEOUT+" seconds to load the perspective " + perspectiveID + ".")
+        //        .isLessThan(PERSPECTIVE_LOADING_TIMEOUT);
+    }
+
+    private void verifyPerspectiveIsLoaded(String perspectiveID, By elementInPerspective) {
+        try {
+            Waits.elementPresent(elementInPerspective, 10);
+        } catch (WebDriverException ex) {
+            Assertions.fail("The perspective with ID " + perspectiveID + " could not be loaded.");
```
jhrcek: I forgot we should add `ex` as the second parameter of `Assertions.fail` not to loose the failure stacktrace. Please add it.

kie-wb-tests/kie-wb-tests-gui/src/test/java/org/kie/wb/selenium/ui/StandalonePerspectivesIntegrationTest.java
triceo: OK.

kie-wb-tests/kie-wb-tests-gui/src/main/java/org/kie/wb/selenium/util/ApplicationLoadingPopup.java
```diff
@@ -0,0 +1,36 @@
+package org.kie.wb.selenium.util;
```
manstis: License please

kie-wb-tests/kie-wb-tests-gui/src/main/java/org/kie/wb/selenium/util/PerspectiveFactory.java
```diff
@@ -0,0 +1,17 @@
+package org.kie.wb.selenium.util;
```
manstis: License please

kie-wb-tests/kie-wb-tests-gui/src/test/java/org/kie/wb/selenium/ui/StandalonePerspectivesIntegrationTest.java
```diff
@@ -0,0 +1,121 @@
+package org.kie.wb.selenium.ui;
```
manstis: License please

kie-wb-tests/kie-wb-tests-gui/src/test/java/org/kie/wb/selenium/ui/LoadAllPerspectivesIntegrationTest.java
```diff
@@ -31,37 +30,26 @@
...
         HomePerspective home = login.loginDefaultUser();
 
-        for ( Persp<?> p : Persp.getAllPerspectives() ) {
-            if ( perspectiveInCurrentWebApp( p ) ) {
-
-                Persp<?> _p = p;
-                LOG.info( "Checking perspective '" + p.getName() + ".." );
-                AbstractPerspective perspective = home.getNavbar().navigateTo( _p );
-
-                //ProjectAuthoring will direct to ProjectLibrary if there are no Projects in the workbench
```
manstis: Why did you remove this? (Might be obvious to me later...)

jhrcek: Just removed the inner if and reformatted the code. The if was necessary when we had the dichotomy Authoring/Library (navigating to authoring with no project redirected to library). Since Authoring has been removed and the "Project Authoring" menu item now always takes you to library perspective, we no longer need this "skip authoring" workaround

kie-wb-tests/kie-wb-tests-gui/src/test/java/org/kie/wb/selenium/ui/StandalonePerspectivesIntegrationTest.java
kkufova: Good point, moved after `ApplicationLoadingPopup`.

Rikkola:
> Asset Management: AFAIK this feature was removed in 7.x. Shouldn't this perspective be removed as a part of a cleanup?

I'm just removing Asset Management with upcoming Project Structure changes. 

> AuthoringPerspective and AuthoringPerspectiveNoContext: Are these two perspectives still relevant now that we have the Library?

They are. I tested these for 7.0-final. They still make sense to have in, but the navigation in the perspective is more limited. Can't change repositories, branches or projects, but you can navigate inside the project.

kie-wb-tests/kie-wb-tests-gui/src/test/java/org/kie/wb/selenium/ui/StandalonePerspectivesIntegrationTest.java
kkufova: Done.

jhrcek:
@Rikkola regarding `AuthoringPerspective` vs. `AuthoringPerspectiveNoContext`. I remember vaguely that the only difference was that in the latter it was not possible to navigate to different projects. Now that Project explorer no longer has controls to do that kind of navigation, wouldn't it make sense to remove `AuthoringPerspectiveNoContext` altogether? Or do these 2 differ in some other respect?

Rikkola:
@jhrcek I think there is still some small difference, but now your making me doubt what I remember. Maybe I am wrong.

kie-wb-tests/kie-wb-tests-gui/src/main/java/org/kie/wb/selenium/util/ApplicationLoadingPopup.java
kkufova: Done.

kie-wb-tests/kie-wb-tests-gui/src/main/java/org/kie/wb/selenium/util/PerspectiveFactory.java
kkufova: Done.

kie-wb-tests/kie-wb-tests-gui/src/test/java/org/kie/wb/selenium/ui/StandalonePerspectivesIntegrationTest.java
kkufova: Done.

paulovmr:
Hi, details about the new home page are [here](https://issues.jboss.org/browse/BPMSPL-16).

About the `PreferencesCentralPerspective`, it can only be accessed with some parameters that, for instance, the Admin page sets when the preferences are accessed.

jhrcek:
@manstis @Rikkola @pefernan @paulovmr @kkufova Here's a summary I distilled from the comments so far. Please check it and comment if you disagree about anything / can advise about unresolved items. I'll keep this updated based on your comments.

## UNRESOLVED
Q: Which perspectives do we want to test in this way?
A: Probably only AuthoringPerspective and AuthoringPerspectiveNoContext as this is the primary use case for standalone perspectives.
Testing other perspectives might make this test unnecessarily flaky because:
 - other perspectives will be added/removed over time
 - most of them are not designed for standalone use (?) / require additional parameters to be used (e.g. PreferencesCentralPerspective)
 - Issue 2. below

2. Do we want to investigate "Loading time of some standalone perspectives sometime exceeds 30seconds" issue?

3. AuthoringPerspective vs. AuthoringPerspectiveNoContext
Since Project explorer no longer has OrgUnit/Repo/Project navigation the distinction between these 2 might no longer be relevant. Do we want to keep both?

## RESOLVED
Q: Do we want to remove some perspectives from the workbench?
A: Filed [GUVNOR-3323](https://issues.jboss.org/browse/GUVNOR-3323) to remove `WiresGridsDemoPerspective` and `WiresTreesPerspective`, Toni is removing `Asset Management`.

Q: Do we want to test more stuff than that the perspective was loaded?
A: Since we'll probably focus on authoring only, we'll also test opening some asset, checking it was loaded and effects of additional parameters (enabling/disabling build buttons).

pefernan:
@jhrcek regarding FormDisplayPerspective: I think that this perspective should be removed... it was used to integrate forms with remote apps using the old Rest services

jhrcek:
@pefernan Ok, thanks. I filed [JBPM-6231](https://issues.jboss.org/browse/JBPM-6231) to track the removal of that and assigned to you. Feel free to reassign as needed.

manstis:
Thanks for the summary @jhrcek.

I think testing only ```AuthoringPerspective``` and ```AuthoringPerspectiveNoContext``` would be fine (these are the only perspectives we've ever had JIRAs for in "standalone" mode). I think ```AuthoringPerspectiveNoContext``` excludes the Project Explorer all together so probably needs to be kept (as ```AuthoringPerspective``` has the Project Explorer in a "dock" IIRC).

pefernan:
@jhrcek thanks!

jhrcek:
@manstis either you don't remember it correctly or we have a regression as the following screenshot illustrates. Which is the case? Your answer will resolve our last unresolved concern and we'll refactor the proposed test to reflect the discussion.

authoringnocontexthasprojectexplorer

Steps I used:
1) in "standard" mode (NON standalone) go to library and import  evaluation project to have some project navigable in project explorer.
2) Navigate to url: http://localhost:8080/business-central/kie-wb.jsp?standalone&perspective=AuthoringPerspectiveNoContext

Actual result: Project explorer is shown and I'm able to open any asset in the project.
Expected result - P.ex. is not shown

Rikkola:
@manstis, @jhrcek, everyone.
Ok, did some investigation and managed to find the original specification (one liners). In this case they help a lot. [RHBPMS-509](https://issues.jboss.org/browse/RHBPMS-509)

So looks like there should not be a difference between the two perspectives.

manstis:
@jhrcek @Rikkola Thanks; I may be mistaken.. 

That said looking at the current [code](https://github.com/kiegroup/kie-wb-distributions/tree/master/kie-drools-wb-parent/kie-drools-wb-webapp/src/main/java/org/kie/workbench/drools/client/perspectives) I can see that ```AuthoringPerspective``` **always** adds Project Explorer in a "dock" whereas ```AuthoringPerspectiveNoContext``` has the ability to provide some configuration. 

I've not looked further into what these options actually _do_ (or should do!)

Rikkola:
@manstis Well, that is what the original specification says. However I fear that how it actually worked might have been different.

The options indeed might be different. I have to admit I thought `AuthoringPerspective` also has them all.

jhrcek:
@manstis @Rikkola thanks for filling in all the missing pieces! We'll use the original (RHBPMS-509)[https://issues.jboss.org/browse/RHBPMS-509] jira as a "specification" to implement the test(s) (taking into configuration recent removal of project context dropdowns of course).

kkufova:
@manstis @Rikkola @jhrcek 
I looked into all the available resources and manually tested both perspectives. What I discovered is that `AuthoringPerspective` and `AuthoringPerspectiveNoContext` are almost identical. There are only two minor differences:
1. parameter `header=AppNavBar` is valid for both perspectives, however, the navigation bar differs. I will talk about this later in this comment.
2. parameter `path` is valid only for `AuthoringPerspectiveNoContext`. If you don't specify this parameter, `AuthoringPerspectiveNoContext` acts exactly the same as `AuthoringPerspective`.

Parameter `explorer_mode` is valid for both perspectives and works exactly the same. Two more parameters mentioned in [RHBPMS-509](https://issues.jboss.org/browse/RHBPMS-509) (`no_context_navigation` and `no_search`) are not valid anymore.

Therefore, I think it only makes sense to test `AuthoringPerspectiveNoContext` which is slightly more complex. What is more, I suggest merging these two perspectives or removing one of them. Keeping both does not make sense as any of them offers any extra functionality. @manstis @Rikkola WDYT?

Regarding the navigation bar, in `AuthoringPerspectiveNoContext`, it looks like follows:
prvni
In `AuthoringPerspective`, the navigation bar contains more options:
druhy
However, I don't think all of the options (with the exception of **New Item**) should be available:
- **Example** allows you to import an example project or a project from a custom repository. This doesn't really make sense, as the imported project cannot even be opened when using a standalone Authoring Perspective.
- **Repository** works very similarly as the **Example** described above. What is more, managing the modules in this screen causes Business Central to freeze.
- **Messages** and **Asset Search** work fine, but I don't think they are necessary. If they are, they should be added to `AuthoringPerspectiveNoContext` as well.

If we decide to keep the `AuthoringPerspective`, I think these issues should be fixed.

Rikkola:
Thanks for taking a look @kkufova.
Yes **Example** makes no sense.
I'm about to remove **Repository** for all the menu's.
**Messages** and **Asset Search** should be needed in both I think.

All in all. Now that we have the library screens as the new authoring. We should probably replace these two with access to that perspective. The search for example will be removed from everywhere except the library and in library it will be done with filters in the file list screen.

I propose we ask for the design team to take this into consideration and provide us a wire frame (if one is needed) and we wrap up some sort of specification so that we all now what it should do. Looks like nobody really knew what these two did.

Also ping @paulovmr.

manstis:
@kkufova Could you please advise the status of this PR?

kkufova:
@manstis I guess we can close this PR and I will create a new one after it is clear which standalone perspective should I actually test. Could you please let us know if you agree with creating a brand new Library standalone perspective as @Rikkola suggested? I was waiting for your opinion. I can file a Jira for removing `AuthoringPerspective` and `AuthoringPerspectiveNoContext` perspectives and creating the new one. Should @paulovmr be the assignee of such task?

manstis:
After much kerfuffle it became evident that nobody really knows _where_ the original "standalone" requirement originated; nor _whom_ can comment on what the future should look like now we have "Library" nor whether we need the two perspectives or not. @kkufova has kindly offered to raise the issue of "standalone" perspectives in the upcoming UUAX meeting with @ibek and (possibly) @lclay2

PR #556, opened 2017-07-20
Add swarm plugin to productized profile.

AntStephenson:
@psiroky Could you review? thanks!

### DIFF
--- a/kie-drools-wb-parent/kie-drools-wb-webapp/pom.xml
+++ b/kie-drools-wb-parent/kie-drools-wb-webapp/pom.xml
@@ -2053,6 +2053,31 @@
                 <classifier>redhat</classifier>
               </configuration>
             </plugin>
+            <plugin>
+            <groupId>org.wildfly.swarm</groupId>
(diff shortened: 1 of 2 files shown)


kie-drools-wb-parent/kie-drools-wb-webapp/pom.xml
```diff
@@ -2053,6 +2053,31 @@
...
+              </fractions>
+            </configuration>
+            <executions>
+              <execution>
+                <goals>
+                  <goal>package</goal>
+                </goals>
+              </execution>
+            </executions>
+          </plugin>
```
psiroky: Could we move the config outside of the profiles? That would make sure it would be build even for -Dproductized. And it would save us some duplication.

psiroky:
We also need to make sure to cherry-pick this into master.

psiroky:
The `-Dproductized` build is failing because of https://issues.jboss.org/browse/SWARM-1332

psiroky:
Replaced by https://github.com/kiegroup/kie-wb-distributions/pull/557

PR #557, opened 2017-07-20
Make sure swarm plugin runs with -Dproductized

psiroky:


psiroky:
Just played a bit with the swarm jar. I can login, create a project, do a build. What does not work is process designer. But I guess that will need to be fixed separately as at this point I have no idea what is going on.

AntStephenson:
Looks good to me, thanks Petr!

PR #558, opened 2017-07-25
[GUVNOR-3280] move CSRF property to ErraiService.properties

psiroky:
Thanks for the fix @mbarkley!


psiroky:
@mbarkley can you please approve?

PR #561, opened 2017-07-26
[GUVNOR-3280] move CSRF property to ErraiService.properties

psiroky:
Backport of #558


manstis:
@tiagobento FYI, do you need to make similar changes for ```ErraiService.properties``` (support it being in ```META-INF```)?

tiagobento:
@manstis Good catch! `ErraiApp.properties` in the `META-INF` dir is supported in Errai 4.1.x. Since this PR was merged in 7.2.x (which uses Errai 4.0.x), `ErraiApp.properties` **has** to be in the root dir. If this PR is going to be merged in 8.0.x too, then it's better to move it to `META-INF`.

PR #562, opened 2017-07-31
Extend workbench client creation methods

sutaakar:
To simplify Workbench client usage - users don't need to create model instances.

### DIFF
--- a/kie-wb-tests/kie-wb-tests-rest/src/main/java/org/kie/wb/test/rest/client/WorkbenchClient.java
+++ b/kie-wb-tests/kie-wb-tests-rest/src/main/java/org/kie/wb/test/rest/client/WorkbenchClient.java
@@ -67,6 +67,16 @@ public interface WorkbenchClient {
     CreateOrCloneRepositoryRequest createOrCloneRepository(RepositoryRequest repository);
 
     /**
+     * [POST] /repositories
+     */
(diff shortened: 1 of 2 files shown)


kie-wb-tests/kie-wb-tests-rest/src/main/java/org/kie/wb/test/rest/client/WorkbenchClient.java
```diff
@@ -97,6 +117,21 @@
...
+
+    /**
+     * [POST] /organizationalunits
+     */
+    CreateOrganizationalUnitRequest createOrganizationalUnit(String orgUnitName, String owner, String description);
+
+    /**
+     * [POST] /organizationalunits
+     */
+    CreateOrganizationalUnitRequest createOrganizationalUnit(String orgUnitName, String owner, String description, String groupId);
```
psiroky: This is fine considering the other parts of the interface. Just a wild thought: would it make sense to have just `CreateOrganizationalUnitRequest createOrganizationalUnit(OrganizationalUnit organizationalUnit)` + create a builder for the `OrganizationalUnit` (which would have optional fields)?

sutaakar: It may have a sense.
My intention here was to allow users to use Workbench client without knowledge of its inner parts, without need to instantiate its model.

PR #572, opened 2017-08-04
JBPM-6274 - Update batik to 1.9 for 6.5.x

mswiderski:


mswiderski:
@krisv here is the fix for WAS deployment failure for 6.4.5

PR #585, opened 2017-08-14
Remove double '*' pattern (doesn't work in Tomcat).

mbarkley:
@manstis I'm fairly certain this will fix the build issue with Tomcat caused by my previous PR.

@psiroky Is there a quick way to test this? I don't have Tomcat setup locally.


psiroky:
@mbarkley you should be able to run the kie-wb-tests locally on Tomcat with something like `mvn clean verify -Pkie-wb,tomcat8`

mbarkley:
@psiroky Thanks. Giving that a try now.

mbarkley:
@manstis Yeah the `*.cache.html` pattern just in case we ever switch GWT linkers (one of them generates HTML files with JS in them instead of JS files).

mbarkley:
@manstis Well it still fails locally. This time "/org.kie.workbench.KIEWebapp/*.cache.js" is an invalid url constraint.

I will continue investigating solutions.

mbarkley:
@manstis I've discovered the [error](https://stackoverflow.com/a/26744/3091732) of my ways. I've run the tests locally with Tomcat to verify this change fixes this problem.

PR #578, opened 2017-08-08
GUVNOR-3236: Rearrange menus structure

paulovmr:

PR #593, opened 2017-08-17
JBPM-6148 - New kie-wb-runtime-webapp

cristianonicolai:
New module for war containing only runtime related views.


kie-wb-parent/kie-wb-home-page-community/src/main/java/org/kie/workbench/client/home/AbstractHomeProducer.java
```diff
@@ -0,0 +1,66 @@
...
+import org.kie.workbench.client.resources.i18n.Constants;
+import org.kie.workbench.common.screens.home.model.HomeModel;
+import org.kie.workbench.common.screens.home.model.HomeModelProvider;
+import org.kie.workbench.common.screens.home.model.ModelUtils;
+import org.uberfire.client.mvp.PlaceManager;
+
+import static org.kie.workbench.common.workbench.client.PerspectiveIds.*;
+import static org.uberfire.workbench.model.ActivityResourceType.PERSPECTIVE;
+
+public abstract class AbstractHomeProducer implements HomeModelProvider {
```
bsremac: ```AbstractHomeProducer.java``` is showing up twice in this PR. They look identical though, so it seems only as a case of a duplicate.

kie-wb-parent/kie-wb-runtime-webapp/src/main/java/org/kie/workbench/client/navbar/AppNavBar.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    @Override
+    public String getId() {
+        return "AppNavBar";
+    }
+
+    @Override
+    public int getOrder() {
+        return 20;
```
bsremac: Can we replace this magic number with a constant, let's say ```HEADER_STACKING_ORDER```?

kie-wb-parent/kie-wb-runtime-webapp/src/main/java/org/kie/workbench/client/navigation/NavTreeDefinitions.java
```diff
@@ -0,0 +1,149 @@
...
+    public static final String ENTRY_DASHBOARDS = "wb_entry_dashboards";
+
+    public static final String GROUP_DEVOPS = "wb_group_devops";
+    public static final String ENTRY_DEPLOYMENTS = "wb_entry_deployments";
+    public static final String ENTRY_EXECUTION_SERVERS = "wb_entry_execution_servers";
+
+    public static final String GROUP_MANAGE = "wb_group_manage";
+    public static final String ENTRY_PROCESS_DEFINITIONS = "wb_entry_process_definitions";
+    public static final String ENTRY_PROCESS_INSTANCES = "wb_entry_process_instances";
+    public static final String ENTRY_ADMINISTRATION_TASKS = "wb_entry_administration_tasks";
```
bsremac: We had a discussion the other day around proper naming of this perspective in https://github.com/kiegroup/kie-wb-distributions/pull/578.
I'm pasting here Jan's comment:
_"Aministration Tasks" sounds misleading to me, it sounds like some general application administration tasks, while in fact it's "just" admin view of task list -> IMHO Should be renamed to "Task Administration"_
This remark was accepted, so I think it would be good to be consistent with naming across all deliverables.

cristianonicolai:
@psiroky seems like adding a new gwt module to the compilation is trying to allocate way more resources than available (5gb * 3 modules building in parallel). Cannot allocate memory error seems consistent even if I try to run it locally using `-T1C`. Any advice?

psiroky:
@cristianonicolai yeah, I am not sure there is an easy solution to this. The GWT compiled jumped from 2 to 4 apps. That would require like 32GiB of memory. I guess we will need to simply turn of the parallel build.

kie-wb-parent/pom.xml
```diff
@@ -81,6 +249,55 @@
...
+    <profile>
+      <id>fullProfile</id>
+      <activation>
+        <property>
+          <name>full</name>
+        </property>
+      </activation>
+      <properties>
+        <gwt.memory.settings>-Xmx4g -Xms1g -Xss1M</gwt.memory.settings>
+        <gwt.compiler.localWorkers>4</gwt.compiler.localWorkers>
```
psiroky: This should be probably changed to 2 or maybe even leave the default 1.

cristianonicolai: done, removed the custom settings from the fullProfile. reusing the same properties for the different profiles now.

psiroky:
Adding two more GWT apps will increase the build time by at least 20mins to our already very slow build. I know there is probably not much we can do at this point though.

cristianonicolai:
@psiroky maybe give it a go to only use 2 threads instead (-T2) perhaps that can keep the resources usage under control while still working in parallel. Another workaround would be to add a dependency between modules, to enforce running in the gwt modules in sequence. Apart from that, I guess we are only left with turning the parallel build off.

kie-wb-parent/kie-wb-runtime-webapp/src/main/java/org/kie/workbench/client/navbar/AppNavBar.java
cristianonicolai: @bsremac I definitely plan to review some of these duplicated classes after the initial release. Will try to combine most common classes between all different distribution wars.

kie-wb-parent/kie-wb-runtime-webapp/src/main/java/org/kie/workbench/client/navigation/NavTreeDefinitions.java
cristianonicolai: Sure, it makes sense to me, fixed.

bsremac:
+1

cristianonicolai:
rebased and added fix for settings page, depends on https://github.com/kiegroup/kie-wb-common/pull/1084

psiroky:
I see two options when it comes to the parallel/sequential build
1) Limit the number of threads to 2 (so that at most two GWT compilations would happen in parallel) as mentioned by @cristianonicolai 
2) Disable the parallel build and use gwt.compiler.localWorkers=2, so that there would be only single GWT compilation at any point, but it would use 2 workers, which should make it in theory about 30-40% faster.

I'll do some quick tests locally and post the results here.

cristianonicolai:
Depends on https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/523

cristianonicolai:
Depends on https://github.com/kiegroup/kie-jenkins-scripts/pull/57 to adjust the number of parallel threads for the build.

krisv:
jenkins retest this

krisv:
jenkins retest this

krisv:
jenkins retest this

PR #601, opened 2017-08-22
GUVNOR-3423: Navigation Editor should not allow root menu items and dividers

paulovmr:


paulovmr:
Hi @dgutierr , could you take a look at this PR and the [7.3.x](https://github.com/kiegroup/kie-wb-distributions/pull/602) cherry-pick? Thanks!

dgutierr:
@paulovmr The divider creation is disabled across the entire nav tree. That means other nav components supporting dividers (e.g: menu bars) created could not benefit from this feature. In my opinion this is a minor issue. In the future, we can disable the divider creation only for the "Workbench"  node and let it enabled for the rest of the nodes.

PR #605, opened 2017-08-24
JBPM-6148 - New kie-wb-runtime-webapp module

cristianonicolai:
New module for war containing only runtime related views.
New assemblies for distribution


cristianonicolai:
Depends on https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/529

cristianonicolai:
jenkins please retest this

nmirasch:
@cristianonicolai it will be necessary rename runtime to monitoring like at 7.3.x branch or it will be done in a second step?
Another thing I see different from 7 branch here is the user available permissions, Does make sense do the same update on kie-wb-common and let only the ones on workbench and perpectives?

cristianonicolai:
@nmirasch I was planning to open the same PRs merged into 7.3.x for master, so there is still a few changes pending for master to catch up. Changes in here should be consistent with https://github.com/kiegroup/kie-wb-distributions/pull/593. Same goes for the available permissions, that was handled on later PR.

PR #608, opened 2017-08-25
[7.3.x] JBPM-6148 - Add swarm profile to runtime war

cristianonicolai:


kie-wb-parent/pom.xml
```diff
@@ -241,6 +241,31 @@
...
+              <fraction>logging</fraction>
+              <fraction>undertow</fraction>
+              <fraction>cdi</fraction>
+              <fraction>ejb</fraction>
+              <fraction>jaxrs</fraction>
+              <fraction>ee</fraction>
+              <fraction>security</fraction>
+              <fraction>management</fraction>
+              <fraction>datasources</fraction>
+            </fractions>
```
psiroky: Do both WARs need exactly the same fractions? It seems that the runtime/monitoring WAR should not need all of these? Adding more fractions means the uberjar will be bigger. I would opt for keeping the number of fractions to a only those that are needed.

cristianonicolai: @psiroky I do agree with your concern but at this stage, I believe both distributions have the same dependencies.

PR #626, opened 2017-09-27
Add standalone Authoring Perspective test.

kkufova:
Hello @manstis,
This is the standalone Authoring Perspective test we talked about. Could you please have a look? Let me know if there is anything more you would like to test; now the test only checks whether the perspective was loaded and based on the `header` parameter in the URL, the test checks whether the `AppNavBar` is there.
Also, on which branches do you want to have this test?


kkufova:
@manstis 
Done, do we need another review?

PR #632, opened 2017-10-05
RHBRMS-2991 - Add public resources to Tomcat web.xml

cristianonicolai:


kie-drools-wb-parent/kie-drools-wb-distribution-wars/src/main/assembly/tomcat8/WEB-INF/web.xml
```diff
@@ -678,6 +678,16 @@ See http://www.w3.org/TR/SVG/intro.html#MIMEType. -->
     </auth-constraint>
   </security-constraint>
 
+  <!-- public resources -->
```
manstis: I see access to everything under ```/org.kie.workbench.drools.KIEDroolsWebapp``` (for example) is restricted [here](https://github.com/cristianonicolai/kie-wb-distributions/blob/13f68ed5c402c3ce05ce70cb9bd0d257647f83f5/kie-drools-wb-parent/kie-drools-wb-distribution-wars/src/main/assembly/tomcat8/WEB-INF/web.xml#L632). Do you know if it's better to remove this restriction than add the public sub-folders? (The WAR contains many, many more sub-folders under ```/org.kie.workbench.drools.KIEDroolsWebapp``` than you include in the public configuration).

If you've tried and tested this PR; then fair enough.. I just ask :-)

cristianonicolai: @manstis that is the exact intent, allow only those resources that are needed to render the login/logout pages without duplicating it elsewhere, so making public only the necessary files for that. As for the other resources under _/org.kie.workbench.drools.KIEDroolsWebapp_ folder, I believe they should remain protected to only logged in users.

PR #658, opened 2017-11-28
Remove bogus batik-ext dependency

psiroky:
This dependency should not be here. If it is needed it should be
declared by the module that actually needs it (e.g. the
jbpm-designer-backed).


psiroky:
Depends on https://github.com/kiegroup/jbpm-designer/pull/704

PR #660, opened 2017-11-30
Prevent gwtbootstrap3-extras ending up in WARs

jhrcek:
Before this fix gwtbootstrap3-extras jar ended up in WEB-INF/lib of both kie-wb and kie-wb-monitoring (in kie-drools-wb it's already declared as provided so that war doesn't have this problem).

@cristianonicolai can you please check and merge?


cristianonicolai:
thanks @jhrcek

PR #671, opened 2018-01-16
RHDM-271: Disable CMS perspective in RHDM

dgutierr:
Fix + tests.

@manstis Can you review please?


dgutierr:
@manstis you are right. I finally removed both from community and product.

dgutierr:
@manstis For 7.5.x I had to add the changes done in JBPM-6511 in order to fix some entries in the mega menu.

PR #672, opened 2018-01-16
[7.5.x] RHDM-271: Disable CMS perspective in RHDM 

dgutierr:
Removed CMS perspective from both community & product.

Also backported the changes from https://issues.jboss.org/browse/JBPM-6511 as it was mandatory in order to remove some old entries from the mega menu.

@manstis Can you verify please? Also, if everything is fine, would you mind to merge it before today EOD as we'd like to include it in the tomorrow's build. Thanx!


dgutierr:
@manstis Green light!

PR #666, opened 2018-01-10
RHDM-252 Updated branding for RHDM.

AntStephenson:
Updated branding for RHDM, including new logos provided by UX.

PR #682, opened 2018-01-30
RHDM-252 Updated branding for RHDM. (#666)

AntStephenson:
(cherry picked from commit 1b123453bc9aa73990a12866660293c7510c922f)

There was a conflict in kie-drools-wb-parent/kie-drools-wb-home-page-product/src/main/resources/org/kie/workbench/drools/client/resources/i18n/Constants.properties which was trivial to resolve..

@cristianonicolai let's go for round 2? second time lucky? :-)


cristianonicolai:
@AntStephenson seems to work :)

PR #690, opened 2018-02-09
AF-1007: Workbench *.cache.js files still not properly cached, even w…

mbarkley:
…hen cache header filter configured


mbarkley:
Because of how url-patterns are matched, these extension patterns never actually did anything, so I am removing them to avoid confusion in the future.

mbarkley:
@manstis Would you care to review this?

PR #705, opened 2018-03-12
RHDM-478_7.5.x

lanceleverich:
* Updated pom files so that kie-pmml module is included
* Added kie.conf so that kie-pmml version of PMMLCompiler is used when compiling a pmml asset


mareknovotny:
I guess this PR is outdated and we can close it, as the https://issues.jboss.org/browse/RHDM-478 is committed on branch for 7.7.x https://github.com/kiegroup/kie-wb-distributions/commit/da23582b936f0111f5563040038287771934e5dd and newer?

PR #711, opened 2018-03-16
Review WildFly ports to allow tests to run in parallel

cristianonicolai:


jhrcek:
The selenium test failures can be ignored - I submitted another [PR](https://github.com/kiegroup/kie-wb-distributions/pull/712) to fix those..

PR #733, opened 2018-04-12
[RHBPMS-1622] Set property for very short Errai marshaller class name…

ryanzhang:
…s. (#434)


ryanzhang:

ryanzhang:
The test failure seems irrelevant.

PR #736, opened 2018-04-17
Remove @After in UI test to make screenshots useful

jhrcek:
@manstis could you please merge if the tests pass?
Helping @mbiarnes  with selenium failures troubleshooting on 7.5.x on tomcat8.

It's necessary not to use `@After` in selenium tests, because `ScreenshotOnFailure` `TestWatcher` just captures useless screenshot of the login screen when test fails. Also logout after test is unnecessary (that logout works is already tested in `LoginIntegrationTest`).
The added `@Page` annotation ensures that `home` field is initialized in cases when some previous test already logged into the app.


manstis:
@jhrcek Sure, just let me know when this is green.

jhrcek:
Job mysteriously died. jenkins retest this

jhrcek:
@manstis green

PR #742, opened 2018-04-19
[7.7.x] Replace swarm-plugin with cargo-plugin in Kie-wb-tests

jakubschwan:
Replace swarm plugin with cargo plugin in Kie-wb-tests to have unificated configuration of containers.
Also there was test configuration issue on swarm - tests couldn't be executed on swarm.


kie-wb-tests/pom.xml
```diff
@@ -563,86 +437,91 @@
...
           <plugins>
             <plugin>
-              <groupId>org.wildfly.swarm</groupId>
-              <artifactId>wildfly-swarm-plugin</artifactId>
+              <groupId>org.apache.maven.plugins</groupId>
+              <artifactId>maven-dependency-plugin</artifactId>
               <executions>
                 <execution>
-                  <id>start</id>
+                  <id>download-swarm-wb</id>
```
jhrcek: nitpick:
```
if (my understand it right) {
  you're not actually downloading anything, just copying swarm artifact produced by the build to some dedicated "container" folder?
   and this execution id is misleading and should be changed to something like copy-swarm-wb
} else {
   please explain how this setup works :-)
}

```

jhrcek:
@jakubschwan could you please explain the motivation behind this pr (I know you told us yesterday, but would be helpful for people reviewing this). Also I'm not sure they'll be ok with removing of tomcat8 profile. Although RH doesn't support tomcat8 distro, I think community still supports it.

jakubschwan:
@jhrcek good point, I just added back tomcat8 profile and add explain to the first comment.

PR #744, opened 2018-04-24
AF-1136: Error after logging in as role user

tiagobento:
See https://issues.jboss.org/browse/AF-1136

I couldn't reproduce the exact error from the JIRA but this PR fixes the blank screen that appears when logging in as role 'user'.


tiagobento:
@jhrcek Would you mind taking a look at that one?

PR #745, opened 2018-05-02
BAPL-934: integrate Case Mgmt into Workbench

jeremyary:


jeremyary:
@manstis fair, I'll get a PR set up for the bug fixes in kie-wb-common

jeremyary:
Per @mdproctor, preference to disable CM asset by default needs to be included with PR

PR #760, opened 2018-05-16
Prevent occasional failures in LoadAllPerspectivesIntegrationTest

jhrcek:
`LoadAllPerspectivesIntegrationTest` was waiting for `MySpace` breadcrumb to appear after opening pages perspective, but instead there was `Standalone_space` which is created by `StandalonePerspectivesIntegrationTest` in cases when it runs earlier than the former.  Waiting for 'Spaces' link (which should be always present) should make this more robust.

@Rikkola FYI fix as promised in https://github.com/kiegroup/kie-wb-common/pull/1744#issuecomment-389464685


jhrcek:
@manstis green. Please merge for better selenium stability on master.

Rikkola:
@jhrcek Thank you for this.

PR #773, opened 2018-06-27
AF-1226: Improve error handling displaying additional info to the GenericErrorPopup

tiagobento:
Related:

https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/770
https://github.com/errai/errai/pull/350
https://github.com/kiegroup/appformer/pull/404
https://github.com/kiegroup/kie-wb-common/pull/1925
https://github.com/kiegroup/drools-wb/pull/878
https://github.com/kiegroup/jbpm-wb/pull/1162
https://github.com/kiegroup/optaplanner-wb/pull/288
https://github.com/kiegroup/kie-wb-distributions/pull/773

### DIFF
--- a/kie-drools-wb-parent/kie-drools-wb-webapp/src/main/java/org/kie/workbench/drools/client/KieDroolsWorkbenchEntryPoint.java
+++ b/kie-drools-wb-parent/kie-drools-wb-webapp/src/main/java/org/kie/workbench/drools/client/KieDroolsWorkbenchEntryPoint.java
@@ -30,6 +30,7 @@ import org.jboss.errai.ioc.client.api.EntryPoint;
 import org.kie.workbench.common.workbench.client.admin.DefaultAdminPageHelper;
 import org.kie.workbench.common.workbench.client.authz.PermissionTreeSetup;
 import org.kie.workbench.common.workbench.client.entrypoint.DefaultWorkbenchEntryPoint;
+import org.kie.workbench.common.workbench.client.error.DefaultWorkbenchErrorCallback;
 import org.kie.workbench.common.workbench.client.menu.DefaultWorkbenchFeaturesMenusHelper;
 import org.kie.workbench.drools.client.navigation.NavTreeDefinitions;
 import org.uberfire.client.authz.PerspectiveTreeProvider;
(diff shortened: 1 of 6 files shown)


domhanak:
@tiagobento I only see this for 7.7.x is that what we want?

tiagobento:
@domhanak The plan is to port it to master as well, but 7.7.x is a priority. Not sure if it will make it, though.

@jhrcek Thanks for the review, I'm glad you like it :)

As for the bullets you raised:

1. I understand. This had no UX input whatsoever, so some UI/UX decisions might've not been the best, since they were made by me 😆. I was avoiding a "Close" button, so I added this behavior of closing the Popup to both "Ignore" and "Copy error details" (Although there's a small `X` there that actually closes the Popup)

2. I can display a notification telling "Error details were copied to your clipboard" or something similar. Wdyt? (There's a similar behavior on the General Settings tab when copying the Git URL).

3. The main idea behind this new ErrorPopup is not only gather more information, but also to hide unnecessary details that might put users into thinking about something that is not related at what they're doing at all. Exception messages can be very misleading for non-dev people, so I think the best option is simply not to show it.

4. There's a hidden textarea that contains the error information. Do you think it's possible to access the contents of this element from the tests? If so, problem solved, right?

Given my replies, I think the only change that would make sense to be made is adding a notification after copying to the clipboard. How does that sound?

P.S.: It's nice you were able to see the new Popup in action while reproducing a "random" issue :)

tiagobento:
Forgot to mention, but when you click on "Copy error details", the details are logged on the browser's console, so you don't have to actually paste it somewhere to read it.

jhrcek:
@tiagobento 
> I can display a notification telling "Error details were copied to your clipboard" or something similar. Wdyt?

Ok, adding notification sounds good enough. I'd be interested in UX input, on that, but I understand we don't have time for that before the release. Maybe I'd also change the "Copy error details" to button, because similarly to "Ignore" button it's action oriented (do something) and not navigation (take me somewhere). Wdyt?

>The main idea behind this new ErrorPopup is not only gather more information, but also to hide unnecessary details that might put users into thinking about something that is not related at what they're doing at all.

Got it, in that case I agree with leaving it hidden.

> There's a hidden textarea that contains the error information. Do you think it's possible to access the contents of this element from the tests? If so, problem solved, right?

Great, I didn't notice it's there already. Yes, that will make it easily retrievable from selenium.

tiagobento:
@jhrcek	See the changes on https://github.com/kiegroup/kie-wb-common/pull/1925.

jhrcek:
Ok, running PR build to get the war.. 
jenkins retest this

PR #774, opened 2018-07-03
Make jBPM server Distribution for community only

ryanzhang:
This is for fixing the productizedProfile build error. the org.kie.kie-wb wildfly11 war doesn't exist for productized Profile.


ryanzhang:
@mswiderski  Thanks for the review. 
The failed test case seems irrelevant to this code change.  
I am merging this to recover the prod nightly build.

PR #785, opened 2018-07-30
JBPM-7477 - Updates to better align with Docker image

cristianonicolai:
- Use wbadmin user instead of admin to avoid issues with admin group in human task assignments
- Review user roles
- Add MySQL port number parameter to cli script


mswiderski:
@cristianonicolai the change to user requires changes to:
- kie-server-distributions/jbpm-server-distribution/Readme.txt
https://github.com/kiegroup/kie-wb-distributions/blob/master/kie-server-distributions/jbpm-server-distribution/README.txt
- double check if jbpm-website is not referring to the admin user ... I think it should not but to make sure we don't have outdated information after this change

cristianonicolai:
@mswiderski updated README and updates to website will come later, there is no reference to admin user in there at the moment.

PR #791, opened 2018-08-13
Revert and upgrade memory limits

hasys:


jhrcek:
@hasys just curious: what happened with the first fix? It didn't work?

PR #794, opened 2018-08-20
Update StandalonePerspectivesIntegrationTest.

kkufova:
I updated the obsolete standalone perspectives Selenium test. The following is tested:
- the standalone Library perspective without the header,
- the standalone Library perspective with the header (`UberfireBreadcrumbsContainer`),
- the standalone Data Modeler, and
- the standalone Content Management perspective.

@Rikkola, @jhrcek, could you please review?


kie-wb-tests/kie-wb-tests-gui/src/test/java/org/kie/wb/selenium/ui/StandalonePerspectivesIntegrationTest.java
```diff
@@ -33,37 +32,44 @@
...
             PROJECT_URL = "https://github.com/Rikkola/single-project-test-repository.git",
             SPACE_NAME = "Standalone_space",
             PROJECT_NAME = "Standalone_project";
 
-    private final static String // Assemble URL to standalone perspective
+    private final static String
             BASE_URL = System.getProperty("kie.wb.url"),
             APP_CONTEXT = System.getProperty("app.name") + ".jsp",
+            // URL of a standalone perspective:
+            FULL_STANDALONE_PARAMETER = "?standalone=true",
```
jhrcek: Since [the implementation](https://github.com/kiegroup/appformer/blob/bcec76d7204f2f5ae0a13c286a666c3818ecccf7/uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/workbench/Workbench.java#L203) is only looking at the presence of the `standalone` parameter (and completely ignores the value  - true or false), I don't think it's necessary to test "?standalone=true" and "?standalone" separately. I'd just remove the "=true" part to make the test simpler and not to create the false impression that something like "standalone=false" is valid and would disable the feature.

PR #797, opened 2018-08-23
JBPM-7635 - Upgrade to WildFly 13

mswiderski:
depends on
https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/801
https://github.com/kiegroup/droolsjbpm-integration/pull/1550

### DIFF
--- a/kie-server-distributions/jbpm-server-distribution/src/main/server/standalone.xml
+++ b/kie-server-distributions/jbpm-server-distribution/src/main/server/standalone.xml
@@ -1,6 +1,6 @@
-<?xml version='1.0' encoding='UTF-8'?>
+<?xml version="1.0" ?>
 
-<server xmlns="urn:jboss:domain:5.0">
+<server xmlns="urn:jboss:domain:7.0">
     <extensions>
         <extension module="org.jboss.as.clustering.infinispan"/>
         <extension module="org.jboss.as.connector"/>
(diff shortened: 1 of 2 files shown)


mswiderski:
@cristianonicolai here are the changes to single zip - mainly around wildfly itself as it still points to widlfly11 classifier for workbench and case mgmt showcase

cristianonicolai:
jenkins please retest this

kie-server-distributions/jbpm-server-distribution/src/main/server/standalone.xml
```diff
@@ -1,6 +1,6 @@
-<?xml version='1.0' encoding='UTF-8'?>
+<?xml version="1.0" ?>
```
cristianonicolai: @mswiderski I wonder if we could get away from having the entire file and just run a set of .cli commands on top of the clean installation. That could be run as part of the maven build and we would still have the final config shipped. It would at least avoid having to maintain this kind of file.

mswiderski: that's correct assuming we can execute the jboss-cli from within maven build/assembly not sure how to do it though...

mareknovotny:
just a comment here that referenced dependent PR https://github.com/kiegroup/droolsjbpm-integration/pull/1550 is closed. Is that intention? @mswiderski

kie-server-distributions/pom.xml
```diff
@@ -13,7 +13,7 @@
 
   <properties>
     <version.org.postgresql>42.2.1</version.org.postgresql>
-    <version.org.wildfly>11.0.0.Final</version.org.wildfly>
+    <version.org.wildfly>13.0.0.Final</version.org.wildfly>
```
mareknovotny: we have in PR queue PR for Wildfly 14.0.1

Wouldn't be better to synchronize on that version?

mswiderski:
@mareknovotny out of date as there is another set of PRs for wildfly 14 upgrade thus closing

PR #823, opened 2018-10-03
Prevent home page loading timeouts in selenium tests

jhrcek:
The root cause of these timeouts is the time it takes to download and execute the 23MB main workbench javascript file (compiled from GWT).

Even on my overpowered laptop the initial load of the home page takes ~13 seconds, so the default 15s home page loading timeout is unreasonably small.
Bumping the default to 30 seconds and adding log message so we can check later how slow the loading was.


jhrcek:
@manstis I'm glad I'm not the only one who sometimes thinks about outdating docs :smile:

PR #835, opened 2018-10-25
[JBPM-7869] NoClassDefFoundError warnings when starting Workbench

elguardian:


elguardian:
wdyt @Rikkola  ?

elguardian:
ok to test

Rikkola:
Looks good. Thank you.

@jomarko

jomarko:
@elguardian @Rikkola thank you both, will let you know once done with review.

elguardian:
seems 1 there was a jira already there
https://issues.jboss.org/browse/DROOLS-2207

elguardian:
seems there is a class that should be there in runtime
https://github.com/kiegroup/drools-wb/blob/master/drools-wb-services/drools-wb-verifier/drools-wb-verifier-webworker/src/main/java/org/drools/workbench/services/verifier/plugin/backend/DecisionTableVerifierWebWorkerProvider.java

jomarko:
@elguardian ok, sorry, juts to align my understanding of your comment, will you check both **kie-drools-wb** and **kie-wb** or are you saying they should be done separately?

Rikkola:
@elguardian There is some client side code, but that is not accessed by the main client side of the workbench. The client side there is compiled to a separate javascript file that is then used as a webworker.

Rikkola:
So what the webworker module does is compile another GWT application, packages it into the jar and then that jar provides the js for the client side of the main workbench.

elguardian:
@Rikkola (just to double check)
I was referring to this error (that now is thrown with the dependencies set with this PR)

`16:24:54,823 ERROR [org.kie.workbench.common.services.verifier.service.VerifierWebWorkerServlet] (default task-1) Failed to load verifier web worker. Verifier with id /dtableVerifier/dtableVerifier.nocache.js was not found.`

This is thrown by the servlet because I marked the dependecy as provided.

https://github.com/kiegroup/kie-wb-common/blob/master/kie-wb-common-services/kie-wb-common-verifier/kie-wb-common-verifier-service/src/main/java/org/kie/workbench/common/services/verifier/service/VerifierWebWorkerServlet.java#L59

It is not clear to me if this is error is ok or not.

jomarko:
@elguardian it is not ok as it causes Decision Table Verifier is missing in workbench UI.

Rikkola:
@elguardian yeah that means the web worker is not found, but the reason why it is not found might be that the drools-wb was not compiled with -Dfull.

This issue keeps coming up, not sure what I should do to fix it. Nobody expects to run drools-wb with the handle. Rather they skip gwt compile all together.

Rikkola:
Ok. So digging deeper, got the same report from other sources. The DecisionTableVerifierWebWorkerProvider class is needed, but it is also the only class needed in from the drools-wb-verifier-webworker module. 
Needed in the jar that is.
The other classes are needed to generate the JavaScript for the web worker and the webworker needs to be generated in that module. Wondering if Maven let's us leave the other classes our someway.

Rikkola:
@elguardian @jomarko I'm working on a fix. The fix should be to exclude the client side classes from the webworker jar.

elguardian:
@Rikkola 
problably could be done excluding from cdi in the beans.xml

Rikkola:
This should do the trick.

https://github.com/kiegroup/drools-wb/pull/983

PR #842, opened 2018-11-07
Don't use wildfly version in property names

jhrcek:
also reuse `version.org.wildfly` coming from kie-parent


jhrcek:
@manstis green (wb rest and selenium tests influenced by this change passed). Could you please merge this?

PR #843, opened 2018-11-07
updated to new Zanata project

mbiarnes:


mbiarnes:
@manstis there are two test failures - the build was successful - the tests have nothing to do with this upgrade - I would say I can merge

PR #854, opened 2018-12-05
Move jbpm-server-distribution to business-central-parent

cristianonicolai:
Allow artifacts to be available before tests run.


cristianonicolai:
@jsoltes this should help to have the zip file available in the downstream build, independently from the execution os the Selenium tests. It tweaks the maven reactor order.

mswiderski:
@cristianonicolai looks ok to me ... just wondering shall we also update the context path from Jbpm-console to business-central?

cristianonicolai:
@mswiderski yep, makes sense, I would just rather do it in a separated/specific PR.

PR #857, opened 2018-12-11
AF-1655: Adding C3 dependency

jesuino:
Adding the new chart implementation based on C3 as a dependency.


cristianonicolai:
jenkins please retest this

PR #863, opened 2019-01-15
DROOLS-3506 : Enable Test Scenario (Preview) by default

Rikkola:
https://issues.jboss.org/browse/DROOLS-3506

@kkufova @gitgabrio @danielezonca


business-central-parent/business-central-distribution-wars/business-central/src/main/productized/resources/security-policy.properties
```diff
@@ -57,7 +57,7 @@ default.permission.editor.read.GuidedDecisionTreeEditorPresenter=false
 default.permission.editor.read.GuidedScoreCardEditor=false
 default.permission.editor.read.ScoreCardXLSEditor=false
 default.permission.editor.read.BPMNDiagramEditor=true
-default.permission.editor.read.ScenarioSimulationEditor=false
+default.permission.editor.read.ScenarioSimulationEditor=true
```
danielezonca: Is it the same of just remove this line? I'm asking because I cannot find DMN Editor entry in the file and I expect it used the same mechanism.
If yes I would like to just remove instead of keep it as "useless" configuration

business-central-parent/business-central-webapp/src/main/resources/security-policy.properties
```diff
@@ -58,7 +58,7 @@ default.permission.editor.read.GuidedDecisionTreeEditorPresenter=true
 default.permission.editor.read.GuidedScoreCardEditor=true
 default.permission.editor.read.ScoreCardXLSEditor=true
 default.permission.editor.read.BPMNDiagramEditor=true
-default.permission.editor.read.ScenarioSimulationEditor=false
+default.permission.editor.read.ScenarioSimulationEditor=true
```
danielezonca: Same comment

business-central-parent/business-central-distribution-wars/business-central/src/main/productized/resources/security-policy.properties
Rikkola: Ok so it actually works without this. I thought the default would be false, but looks like I was wrong.

Rikkola:
@kiegroup/gatekeepers please merge this one.

PR #874, opened 2019-02-06
AF-1724: Fixing Brand Image alt text

jesuino:
AF-1724: Fixing Brand Image alt text


jesuino:
Would you please check this @ederign @jhrcek I am just fixing the brand image text as found by Jan in  -> https://issues.jboss.org/browse/AF-1724

PR #876, opened 2019-02-12
AF-1798 : Standalone Decision Central package should provide better default configuration

Rikkola:
https://issues.jboss.org/browse/AF-1798

### DIFF
--- a/business-central-parent/business-central-distribution-wars/src/main/shared/swarm/sample-swarm-config.yml
+++ b/business-central-parent/business-central-distribution-wars/src/main/shared/swarm/sample-swarm-config.yml
@@ -1,0 +1,70 @@
+# Sample Swarm config for KIE Drools Workbench
+swarm:
+  management:
+    blocking:
+      timeout: 900 # This might need to be higher if the jar has timeout issues on startup
+    security-realms:
+      ApplicationRealm: # almost the same as default config in WildFly 10.1.0.Final
+        local-authentication:
+          default-user: local
+          allowed-users: local
+          skip-group-loading: true
+        properties-authentication:
+          path: application-users.properties
+          plain-text: true
+        properties-authorization:
+          path: application-roles.properties
+# Optional logging
+  logging:
+    console-handlers:
+      CONSOLE:
+        level: INFO
+        named-formatter: COLOR_PATTERN
+    pattern-formatters:
+      PATTERN:
+        pattern: "%d{yyyy-MM-dd HH:mm:ss,SSS} %-5p [%c] (%t) %s%e%n"
+      COLOR_PATTERN:
+        pattern: "%K{level}%d{HH:mm:ss,SSS} %-5p [%c] (%t) %s%e%n"
+    periodic-rotating-file-handlers:
+      FILE:
+        file:
+          path: server.log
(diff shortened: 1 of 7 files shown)


Rikkola:
@domhanak Hi, your marked as the tester for this.

business-central-parent/business-central-distribution-wars/src/main/shared/swarm/sample-swarm-config.yml
```diff
@@ -0,0 +1,70 @@
...
+  kie:
+    server: # Sample connection to KIE Server.
+      id: kieserver1
+      user: kieserver
+      pwd: kieserver1!
+      location: http://localhost:8230/kie-server/services/rest/server
+      controller:
+        user: kieserver
+        pwd: kieserver1!
+      controller: http://localhost:8080/business-central/rest/controller
```
Rikkola: @domhanak There is a small concern about if this works. Ideally the controller branch should not break, but if I have the url in the higher "controller:" I get an error on startup.

domhanak:
Roses are red,
violets are blue,
ignorance is bliss,
Jenkins retest this

domhanak:
Roses are red,
violets are blue,
Jenkins execute full downstream build!

domhanak:
jenkins retest this

domhanak:
jenkins retest this

Rikkola:
jenkins retest this

Rikkola:
@etirelli @mbiarnes This turned green. It should go in before the tag, but @domhanak has not had time to test yet. It works for me locally and @gitgabrio has also tried a local build. This build has no affect on the other builds, so even if it is bad it can not break anything.

business-central-parent/business-central-distribution-wars/src/main/shared/swarm/sample-swarm-config.yml
```diff
@@ -0,0 +1,70 @@
...
+      security:
+        management:
+          wildfly:
+            cli:
+              user: admin
+  kie:
+    server: # Sample connection to KIE Server.
+      id: kieserver1
+      user: kieserver
+      pwd: kieserver1!
```
domhanak: this does not match the password in `application-users.properties`  - there is `kieserver` as password in mentioned file

business-central-parent/business-central-distribution-wars/src/main/shared/swarm/sample-swarm-config.yml
domhanak: the `controller` tags here do not make any difference and the correct url would be `http://localhost:8080rest/controller`

domhanak: the `controller` tags here do not make any difference and the correct url would be `http://localhost:8080rest/controller`

business-central-parent/business-central-distribution-wars/src/main/shared/swarm/sample-swarm-config.yml
Rikkola: Thanks. I'll make a fix.

PR #879, opened 2019-02-18
AF-1798 : Standalone Decision Central package should provide better default configuration

Rikkola:
https://issues.jboss.org/browse/AF-1798


business-central-parent/business-central-distribution-wars/src/main/shared/swarm/sample-swarm-config.yml
```diff
@@ -58,13 +58,3 @@ org:
           wildfly:
             cli:
               user: admin
-  kie:
-    server: # Sample connection to KIE Server.
-      id: kieserver1
-      user: kieserver
-      pwd: kieserver1!
-      location: http://localhost:8230/kie-server/services/rest/server
```
domhanak: Please add these too - you need them to deploy project on kie-server

PR #888, opened 2019-03-06
AF-1876 : Full profile is the default profile in Swarm RHDM distribution

Rikkola:
https://issues.jboss.org/browse/AF-1876

@tomasdavidorg @mareknovotny if you two could please take a look.


Rikkola:
To test use -Dproducized when building.

PR #890, opened 2019-03-07
RHPAM-1205 Loading application... pop up is garbled in locale=ja

bxf12315:
@cristianonicolai Please review it.


manstis:
This will be override the next time translations are fetched from Zanata. Unfortunately this is a fix for the symptoms  of the cause. Where are the Japanese translations garbled? Following a build with -Dfull?

bxf12315:
@manstis It different to other Constants files, it's using in jsp page.
I find some explanation from  stackoverflow. https://stackoverflow.com/questions/28026342/internationalization-using-resource-bundle-properties-in-jsp-non-latin-text-bec

manstis:
@bxf12315 Hi, if you look at the commit you mentioned (when the file was last updated) you'll see the PR is titled "latest Zanata translation changes in kie-wb-distributions from 2019-01-25".. i.e. the i18n'ed text came from Zanata and hence when this is next performed your changes here will again be overwritten.

The files from Zanata are updated to UTF-8 by [this](https://github.com/kiegroup/kie-wb-distributions/blob/master/pom.xml#L194) however it appears to be missing the specific entries for the "Product Home Page" (and no doubt Community too).

I'd suggest updating the `native2ascii-maven-plugin` plugin configuration and trying a `mvn clean install -Dfull` and check the resulting `Constants_ja.properties` has the correct encoding.

cristianonicolai:
Closing as it has been fixed by https://github.com/kiegroup/kie-wb-distributions/pull/892

PR #892, opened 2019-03-08
RHPAM-1205 - Fix path to LoginConstants files

cristianonicolai:


cristianonicolai:
@bxf12315 @manstis this is probably what is preventing the Zanata sync to work properly. My guess is that from now on sync should be fine with proper directories.

PR #895, opened 2019-03-11
JBPM-8337 - Use basic authentication to avoid web session creation fo…

cristianonicolai:
…r REST and WebSocket connections

Assembly:
https://github.com/kiegroup/kie-wb-distributions/pull/895
https://github.com/kiegroup/kie-wb-common/pull/2529
https://github.com/kiegroup/jbpm-wb/pull/1315

### DIFF
--- a/business-central-parent/business-central-webapp-common/src/main/webapp/login.jsp
+++ b/business-central-parent/business-central-webapp-common/src/main/webapp/login.jsp
@@ -14,32 +14,33 @@
   } catch (Exception e) {
     locale = request.getLocale();
   }
-  SecureHeadersFilter.applyHeaders(request,
-                                   response);
+  SecureHeadersFilter.applyHeaders(request, response);
+
+  String context = pageContext.getServletContext().getContextPath();
 %>
 <i18n:bundle id="bundle" baseName="${login.bundle.name}"
              locale='<%= locale%>' />
 <!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">
 <html xmlns="http://www.w3.org/1999/xhtml">
 <head>
-  <link rel="stylesheet" href="styles/base.css">
-  <link rel="stylesheet" href="styles/forms.css">
-  <link rel="stylesheet" href="styles/login-screen.css">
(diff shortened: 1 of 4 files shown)


cristianonicolai:
@manstis @ederign @porcelli any potential drawback in not using the appformer login service? This actually avoids creating sessions for REST requests as well. It only works for Wildfly/EAP but that I think it is ok.

porcelli:
I honestly don't know, this has been around forever... it looks fine, tho.

including @domhanak @tomasdavidorg @barboras7

tomasdavidorg:
@cristianonicolai Do you know how this involve the SSO integration? We have steps how to setup that: https://github.com/kiegroup/kie-docs/blob/master/doc-content/shared-kie-docs/src/main/asciidoc/Workbench/KeycloakSSOIntegration/KC_SSO_integ-section.adoc

Will the setup work (with some update) after these changes?

cristianonicolai:
@tomasdavidorg I would expect it to work mainly because the changes I'm proposing here take th.e same steps as listed in the docs. The only real difference I can see is the use of _rest-all_ role. That should probably be the default here, instead of any role. @porcelli wdyt?

cristianonicolai:
@Rikkola would you be able to check if this has any impact in swarm somehow?

cristianonicolai:
@manstis any concerns from your side regarding this change?

manstis:
@cristianonicolai TBH I can only guess at the affect this could have elsewhere.

However given the comments and approval by others I have no objection.

Rikkola:
@cristianonicolai I'll need to build the Swarm jar locally and even then I will only be able to test the default login. I'll compile it tonight.

cristianonicolai:
@manstis @Rikkola thanks

cristianonicolai:
@mswiderski fyi, this will require that _kieserver_ user has at least one valid role in BC for the connection with the controller to work.

cristianonicolai:
@Rikkola thanks

sutaakar:
@errantepiphany This PR probably affects also Workbench images. Changes in web.xml need to be reflected on the image script side.

errantepiphany:
@sutaakar Thanks so much for this. We were actually just noticing some weird stuff with websockets and auth yesterday, so this is definitely something we should review as it sounds related.

cristianonicolai:
@errantepiphany let me know if you need any help/details about this change.

PR #932, opened 2019-05-24
JBPM-8337 - Secure WebSockets connection using Basic auth to avoid cr…

cristianonicolai:
…eating http session


cristianonicolai:
@sutaakar @jstastny-cz @jakubschwan can you please help verify this PR, there should be no side effects apart from adding _rest-all_ role to any user that needs to connect to BC controller using WebSockets.

sutaakar:
Seems to work correctly for Kie server controller.

PR #941, opened 2019-06-19
BXMSPROD-33 add wildfly zip assembly and remove the assembly-kie-wb-eap-7.xml file and its references

Ginxo:


mareknovotny:
Jenkins retest this please

mareknovotny:
@domhanak can you review now the change?

domhanak:
@mareknovotny yes, I will get to it after CR1

jomarko:
@mareknovotny Hi, I think we should do this "cleanup" also on `business-central-tests/pom.xml`, what do you think? There seems to be a lot of properties to enable testing on eap7.

PR #943, opened 2019-06-28
BAQE-993: Clean up ProjectAccessIntegrationTest

jomarko:
This PR simplify the ProjectAccessIntegrationTest class by
- Resuing method for project creation from parent class
- Reusing one project for multiple tests where the test has read only effect


jomarko:
@mareknovotny would you mind to have a look?

PR #946, opened 2019-07-23
AF-2118: Adding uberfire-security-client-backend

jesuino:
brings back security to business-central by adding uberfire-security-client-backend dependency.


jesuino:

PR #989, opened 2019-11-22
RHPAM-2532: changing failing test to use ssh instead git protocol

porcelli:


business-central-tests/pom.xml
```diff
@@ -290,7 +292,7 @@
                 <configuration>
                   <properties>
                     <cargo.jboss.configuration>standalone</cargo.jboss.configuration>
-                    <cargo.jvmargs>-Xmx1024m</cargo.jvmargs>
+                    <cargo.jvmargs>-Xmx1024m -Dorg.uberfire.nio.git.daemon.enabled=true</cargo.jvmargs>
```
porcelli: for an unknown reason, it only works if I set it here, the above systemProperties doesn't work

business-central-tests/pom.xml
```diff
@@ -158,9 +158,10 @@
               <kie.wb.url>http://${container.host}:${container.port}/${deployable.context}</kie.wb.url>
               <kie.controller.url>ws://${container.host}:${container.port}/${deployable.context}/websocket/controller</kie.controller.url>
               <org.uberfire.nio.git.daemon.port>${cargo.git.port}</org.uberfire.nio.git.daemon.port>
+              <org.uberfire.nio.git.daemon.enabled>true</org.uberfire.nio.git.daemon.enabled>
             </systemPropertyVariables>
             <failIfNoTests>false</failIfNoTests>
-            <argLine>-Xmx1024m -Dfile.encoding=UTF-8</argLine>
+            <argLine>-Xmx1024m -Dfile.encoding=UTF-8 -Dorg.uberfire.nio.git.daemon.enabled=true</argLine>
```
porcelli: for an unknown reason, it only works if I set it here, the above systemProperties doesn't work

paulovmr:
Unrelated test failure.

tomasdavidorg:
@porcelli Why didn't you use the http protocol? The fix is not ideal since the git protocol is disabled by default.

PR #995, opened 2019-11-26
RHPAM-2529: scope provided added to org.jboss.errai:errai-bus dependency from business-central-webapp project

Ginxo:


Ginxo:
wrong PR

PR #998, opened 2019-12-05
Make StandalonePerspectivesIntegrationTest independent of other runs

tomasdavidorg:
There is a mechanism which reruns Selenium tests when some failure
occurs. Sometimes it happens that the first run of first test fails
because of the loading pop-up dialog is still on the screen after login.
Then the second run of the the whole test class fails because of the
space is already created.

This fix creates unique space name for every test run. It could be
possible to just remove the space after the test class but then there
have to be a minute pause before the space is physically deleted.

### DIFF
--- a/business-central-tests/business-central-tests-gui/src/test/java/org/kie/wb/selenium/ui/StandalonePerspectivesIntegrationTest.java
+++ b/business-central-tests/business-central-tests-gui/src/test/java/org/kie/wb/selenium/ui/StandalonePerspectivesIntegrationTest.java
@@ -18,6 +18,7 @@ package org.kie.wb.selenium.ui;
 import org.assertj.core.api.Assertions;
 import org.guvnor.rest.client.CloneProjectRequest;
 import org.guvnor.rest.client.Space;
+import org.junit.AfterClass;
 import org.junit.Before;
 import org.junit.BeforeClass;
 import org.junit.Test;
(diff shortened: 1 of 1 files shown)


tomasdavidorg:

tomasdavidorg:
@hasys Yes. The test can fail. The fail is recorded only in the console log but not in the results. Maybe we can increase the timeout or we should find the cause of the slowness. It might be related to the cargo jvm limits or to the whole build and its memory consumption because locally the test runs fast.

This fix is just isolation of the test from other runs, so we can monitor how stable it is.

tomasdavidorg:
@hasys If you take a look to the console log 
```
09:57:51 [WARNING] org.kie.wb.selenium.ui.StandalonePerspectivesIntegrationTest.testStandaloneEditorPerspective(org.kie.wb.selenium.ui.StandalonePerspectivesIntegrationTest)
09:57:51 [ERROR]   Run 1: StandalonePerspectivesIntegrationTest.testStandaloneEditorPerspective:119->verifyPerspectiveIsLoaded:134 The standalone perspective was not loaded.
09:57:51 [ERROR]   Run 2: StandalonePerspectivesIntegrationTest.testStandaloneEditorPerspective:119->verifyPerspectiveIsLoaded:134 The standalone perspective was not loaded.
09:57:51 [INFO]   Run 3: PASS
```
The test failed twice. The third run was successful. @domhanak is going to prepare some jiras for the slowness.

hasys:
ok, thank you @tomasdavidorg, since there will be more efforts in separate PR to stabilize those tests, this one looks good to me, great fix!

tomasdavidorg:
@manstis @karreiro Could you please take a look or assign the pr review to someone? Thanks. :slightly_smiling_face:

tomasdavidorg:
jenkins retest this

jomarko:
jenkins please retest this

manstis:
@tomasdavidorg Did you want this merged or are you waiting for anyone else?

tomasdavidorg:
Please merge. Thanks @manstis .

PR #1048, opened 2020-08-03
AF-2577: Adding External Component Servlet

jesuino:
Add components servlet and dependency to Business Central

Part of an ensemble, merge with:

kiegroup/droolsjbpm-build-bootstrap#1419
https://github.com/kiegroup/appformer/pull/1016

### DIFF
--- a/business-central-parent/business-central-webapp/src/main/webapp/WEB-INF/web.xml
+++ b/business-central-parent/business-central-webapp/src/main/webapp/WEB-INF/web.xml
@@ -115,8 +115,7 @@
     <url-pattern>/login</url-pattern>
   </servlet-mapping>
 
-	<!-- Custom Component assets server -->
-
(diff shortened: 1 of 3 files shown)


business-central-parent/business-central-webapp/src/main/webapp/WEB-INF/web.xml
```diff
@@ -115,6 +115,17 @@
     <url-pattern>/login</url-pattern>
   </servlet-mapping>
 
+	<!-- Custom Component assets server -->
```
caponetto, line 118:
```suggestion
  <!-- Custom Component assets server -->
```

jesuino: just fixed on latest commit, thanks!

PR #1071, opened 2020-11-23
KOGITO-3891: Stunner - Editors refactoring

romartin:
It **DEPENDS ON**:

- https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1537
- https://github.com/kiegroup/kie-wb-common/pull/3498
- https://github.com/kiegroup/drools-wb/pull/1445
- https://github.com/kiegroup/jbpm-wb/pull/1472

Thanks!

### DIFF
--- a/business-central-parent/business-central-webapp/pom.xml
+++ b/business-central-parent/business-central-webapp/pom.xml
@@ -2053,17 +2053,6 @@
       <scope>provided</scope>
     </dependency>
 
-    <dependency>
-      <groupId>org.kie.workbench.stunner</groupId>
-      <artifactId>kie-wb-common-stunner-kogito-api</artifactId>
-    </dependency>
-
-    <dependency>
-      <groupId>org.kie.workbench.stunner</groupId>
-      <artifactId>kie-wb-common-stunner-kogito-client</artifactId>
-      <scope>provided</scope>
-    </dependency>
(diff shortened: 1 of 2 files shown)


business-central-parent/business-central-webapp/pom.xml
```diff
@@ -2770,8 +2759,6 @@
 
             <!-- TODO kogito needs refactoring out of Business Central -->
             <compileSourcesArtifact>org.kie.workbench:kie-wb-common-kogito-client</compileSourcesArtifact>
```
domhanak, line 2761: Why can't we remove this too?

romartin: yeah @domhanak good catch :) I really wanted to also try to drop this module, but this module is not inside stunner codebase itself (it's part of the repository common modules, see https://github.com/kiegroup/kie-wb-common/tree/master/kie-wb-common-kogito), and once trying to drop it locally I found other usages as well.

So finally I decided not to add more complexity to this PR, otherwise the refactor would take longer and would result even bigger. I think it can be done in a saf and easier way in a second iteration.

Anyway good catch, it's somethign I've also in mind continuously :)

Thanks!

domhanak (approved):

